PR #6, opened 2015-05-20
Raise on missing assigns

chrismccord:
ref: https://github.com/phoenixframework/phoenix/issues/879

@josevalim I expected this to be more work. I am post walking the body and rewriting `Dict.get(var!(assigns), var)` to `Dict.fetch!(var!(assigns), var)`. Do we need anything more? Are there edgecases I'm missing? Should we raise our own more helpful error?

### DIFF
--- a/lib/phoenix_html/engine.ex
+++ b/lib/phoenix_html/engine.ex
@@ -7,7 +7,19 @@ defmodule Phoenix.HTML.Engine do
   use EEx.Engine
 
   @doc false
-  def handle_body(body), do: body
+  def handle_body(body) do
+    Macro.postwalk(body, &postwalk_body/1)
+  end
+  defp postwalk_body({{:., [line: line], [{:__aliases__, _, [:Dict]}, :get]}, _,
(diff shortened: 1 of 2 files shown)


josevalim:
Actually, the original code is here:

https://github.com/phoenixframework/phoenix_html/blob/cm-raise-on-missing-assign/lib/phoenix_html/engine.ex#L84

So instead of post-walking the post-walk, I would simply replace the first post-walk. :) Here is the original code:

https://github.com/elixir-lang/elixir/blob/master/lib/eex/lib/eex/engine.ex#L55-L74

Also, we should consider not using Dict.fetch!, instead use Dict.fetch and raise a clearer error message, like "assign @foo not available in eex template".

Thoughts?

chrismccord:
Ah perfect

josevalim:
:shipit: We should probably release 1.1 after this.

chrismccord:
I'll bump the changelog and release 1.1 over lunch. Thanks!

PR #7, opened 2015-05-20
Clearer error message when 2nd arg to link/2 is not a keyword list

gjaldon:
When calling link/2 like `link("Logout", "/logout")`, you get the ff error `no function clause matching in Keyword.pop/3`. This PR just raises a clearer error message when the 2nd arg to `link/2` is not a keyword list.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -45,23 +45,24 @@ defmodule Phoenix.HTML.Link do
   end
 
   def link(text, opts) do
-    {to, opts} = Keyword.pop(opts, :to)
-  rescue
-    _ -> raise ArgumentError, "link/2 requires a keyword list as second argument"
-  else
-    {nil, _opts} ->
-      raise ArgumentError, "option :to is required in link/2"
-    {to, opts}   ->
-      {method, opts} = Keyword.pop(opts, :method, :get)
+    cond do
+      not match?([{_key, _value}|_opts], opts) ->
+        raise ArgumentError, "link/2 requires a keyword list as second argument"
+      Keyword.get(opts, :to) ->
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/link.ex
```diff
@@ -46,20 +46,22 @@ defmodule Phoenix.HTML.Link do
 
   def link(text, opts) do
     {to, opts} = Keyword.pop(opts, :to)
-    {method, opts} = Keyword.pop(opts, :method, :get)
-
-    unless to do
+  rescue
+    _ -> raise ArgumentError, "link/2 requires a keyword list as second argument"
```
josevalim: Let's not rescue if we don't need to as it can hide other stuff. Instead, let's check explicitly.

gjaldon:
Replaced the `rescue` with a match to check if 2nd arg is a keyword list.

lib/phoenix_html/link.ex
```diff
@@ -45,20 +45,23 @@ defmodule Phoenix.HTML.Link do
...
-
-    if method == :get do
-      content_tag(:a, text, [href: to] ++ opts)
-    else
-      {form, opts} = form_options(opts, method, "link")
-      form_tag(to, form) do
-        content_tag(:a, text, [href: "#", onclick: "this.parentNode.submit(); return false;"] ++ opts)
-      end
+    cond do
+      not match?([{_key, _value}|_opts], opts) ->
```
josevalim: Wouldn't it be better to have a `not is_list/1` guard as another function clause above the current one?

gjaldon: What is the `not is_list/1` guard for?

PR #8, opened 2015-05-22
Merge tags if you want to output tags side by side

gjaldon:
Just thought it'd be nice to have a helper to output tags side-by-side instead of nested


josevalim:
You can just return them in a list. You don't need to do anything else. :)

gjaldon:
Oops! :sweat_smile:  I'm closing this then. :smile:

PR #12, opened 2015-06-02
Make sure nil params submitted to the form are not ignored in favour of their original values

### LINKED ISSUE #11: Nil params are ignored on form re-render
Due to this line here:

https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/form.ex#L805

If an empty value is submitted, scrubbed to nil, and then the form is re-rendered, the nil value is ignored due to it's falsiness and the original value used instead.

Presume it should be like "if the field exists in the params, use it, else go back to the original value".

sevenseacat:
Fixes #11

`Map.get/3` has a third option of a default if the key isn't found in `params`, which seems like exactly what we want.

edit: actually I think the fix is right, but the test is wrong. Looking into it more.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -802,7 +802,7 @@ defmodule Phoenix.HTML.Form do
   ## Helpers
 
   defp value_from(%{model: model, params: params}, field),
-    do: Map.get(params, Atom.to_string(field)) || Map.get(model, field)
+    do: Map.get(params, Atom.to_string(field), Map.get(model, field))
   defp value_from(name, _field) when is_atom(name),
     do: nil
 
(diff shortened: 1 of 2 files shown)


josevalim:
Fetch would be better only because we would never evaluate the other side. But I think the cost of evaluating the other side is probably irrelevant here. :)

sevenseacat:
I'm really stuck on how to test this. `Plug.Test.conn/4` doesn't seem to accept enough options to pass in the persisted data to test against. 

For fetch, do you mean something like the following:

```elixir
handle_dict = fn ->
  {:ok, value} -> value
  :error       -> Map.get(model, field)
end
handle_dict.(Map.fetch(params, Atom.to_string(field)))
```

Feel free to just close this PR and fix it properly :) I tried.

josevalim:
FWIW, your patch looks perfect, I am not sure why it doesn't work. I will try tonight and report back. :)

josevalim:
The issue was that the model was never really set for those models. I am storing a value now before proceeding:

https://github.com/phoenixframework/phoenix_html/blob/master/test/phoenix_html/form_test.exs#L87-L88

Your fix has been merged though and then adapted to use fetch (which is a tiny bit faster), thank you!

PR #15, opened 2015-06-25
Allow default text content on labels to be inferred from field

orendon:
I believe is pretty common to have label text content the same as the field name.
Example, I keep doing this a lot:
```erb
<%= text_input f, :first_name,  "First name" %>
<%= text_input f, :middle_name, "Middle name" %>
<%= text_input f, :last_name,   "Last name" %>
<%= text_input f, :gender,      "Gender" %>
```

So I thought it was a good idea to improve the helper so the label text content can be inferred from the actual field.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -784,6 +784,7 @@ defmodule Phoenix.HTML.Form do
   All given options are forwarded to the underlying tag.
   A default value is provided for `for` attribute but can
   be overriden if you pass a value to the `for` option.
+  Text content would be inferred from `field` if not specified.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


josevalim:

orendon:
@chrismccord @josevalim I'm planning to add `do..end` support to submit/reset inputs (in such case it will render a button instead of input), so we can generate stuff like:

```html
<button type="submit" class="btn btn-success"><i class="fa fa-check"></i> Submit</button>
<button type="reset" class="btn btn-danger"><i class="fa fa-repeat"></i> Reset</button>
```

Wanted to ask before working on the PR, would that be desirable? thx

josevalim:
@orendon that's a good question. Shouldn't we then have a button thing? I.e. should we keep submit and reset as inputs or should we emit whatever makes more sense based on the user input? We typically prefer explicit in Phoenix.

orendon:
@josevalim Agreed on explicitly, I believe a button helper with do..end support will succeed across all scenarios, the type:submit|reset can be passed as function `opts` arguments when needed.

I'm willing to work on a PR to add button helper with do..end support if that sounds ok to you?

josevalim:
Btw, we already have a button thing: https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/link.ex#L111 maybe it is a matter of extending it.

PR #17, opened 2015-07-13
add multiple select support

aarondufall:
Add the option to use multiple select form with the select helper

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -552,9 +552,64 @@ defmodule Phoenix.HTML.Form do
     end
   end
 
+  defp option(option_key, option_value, value_list, acc) when is_list(value_list) do
+    opts = [value: option_value, selected: option_value in value_list]
+    html_escape [acc|content_tag(:option, option_key, opts)]
+  end
+
   defp option(option_key, option_value, value, acc) do
     opts = [value: option_value, selected: value == option_value]
     html_escape [acc|content_tag(:option, option_key, opts)]
+  end
+
+  @doc """
+  Generates a select tag with the given `values`.
+
+  Values are expected to be an Enumerable containing two-item tuples
+  (like maps and keyword lists) or any Enumerable where the element
+  will be used both as key and value for the generated select.
+
+  ## Examples
+
+      # Assuming form contains a User model
+      multiple_select(form, :roles, ["Admin": 1, "Power User": 2])
+      #=> <select id="user_roles" name="user[roles][]">
+          <option value="1">Admin</option>
+          <option value="2">Power User</option>
+          </select>
+
+
+      multiple_select(form, :role, ["Admin": 1, "Power User": 2], values: [1])
+      #=> <select id="user_role" name="user[role]">
+          <option value="1" selected="selected" >Admin</option>
+          <option value="2">Power User</option>
+          </select>
+
+  ## Options
+
+    * `:value` - an Enum of values used to select given options.
+
+    * `:default` - the default value to use when none was given in
+      `:values` and none was available in the model
+
+  All other options are forwarded to the underlying HTML tag.
+  """
+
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -520,6 +520,8 @@ defmodule Phoenix.HTML.Form do
     * `:default` - the default value to use when none was given in
       `:value` and none was available in the model
 
+    * `:mulitple` - set to `true` for multiple select
+
```
wsmoak: oops, small typo, s/mulitple/multiple/

lib/phoenix_html/form.ex
```diff
@@ -531,10 +533,14 @@ defmodule Phoenix.HTML.Form do
...
-      opts
-      |> Keyword.put_new(:id, id_from(form, field))
-      |> Keyword.put_new(:name, name_from(form, field))
+    opts = Keyword.put_new(opts, :id, id_from(form, field))
+      
+    opts = case Keyword.pop(opts, :multiple) do
+        { nil, opts }  -> Keyword.put_new(opts, :name, name_from(form, field))
+        { true, opts } -> 
+          Keyword.put_new(opts, :multiple, "") 
+          |> Keyword.put_new(:name, multiple_name_from(form, field))
```
josevalim: Shouldn't you instead call `name_from(...)` and append `"[]"`?

lib/phoenix_html/form.ex
```diff
@@ -531,10 +533,14 @@ defmodule Phoenix.HTML.Form do
       {prompt, opts} -> {content_tag(:option, prompt, value: ""), opts}
     end
 
-    opts =
-      opts
-      |> Keyword.put_new(:id, id_from(form, field))
-      |> Keyword.put_new(:name, name_from(form, field))
+    opts = Keyword.put_new(opts, :id, id_from(form, field))
+      
+    opts = case Keyword.pop(opts, :multiple) do
```
josevalim: Please indent this as (note the lack of spaces after `{` and `}`):

    opts =
      case Keyword.get(opts, :multiple, false) do
        false -> ...
        true -> ...
      end
      
Notice you don't need to pop the value because Phoenix will do the proper thing if it is true (keep it) or false (remove it).

josevalim:
If we have a multiple select, then the values in the model should be a list, not a single value. So the logic that detects what has been selected needs to be different to. We can either do it behind the scenes or have an explicit multiple_select (as it expects different arguments, it is likely reasonable to have another function).

wsmoak:
see phoenix_html pull #17

aarondufall:
Thanks for the feedback. If it were to be handled behind the scenes, would it work something like: 

* if the field was the same name as the has_many association, it would pull out all the ids on each associated record and place them into a list. that list could then be used to match with ids in the values keyword list for the selected items? 

I agree this could really be is own function I would be happy to move it over if this is the preferred api.

aarondufall:
I have been working on a solution that would use a separate `multiple_select/4` function. After a little reflection and investigation, I noticed that you need to explicitly request for the has_many association to be preloaded. As this may not be the standard way the data is provided to the view, I thought it may work better to have a list of ids provided on a `:values` options. Let me know your thoughts?

josevalim:
@aarondufall Yes, I was thinking something similar. We should just expect the value, regardless if we read it from the model or if it is passed as an option, to be an enumerable. I would still keep the option named `:value` though. So it works similarly to select, except when checking if something was selected or not, we will do `option_value in value` (instead of equality).

aarondufall:
@josevalim Third try's a charm. I believe that I have got the discussed functionality implemented. Any thoughts or feedback is welcome.

josevalim:
Indeed, this is beautiful, thank you!

josevalim:

henrik:
@aarondufall Thank you for this 💚🍰

PR #27, opened 2015-09-01
Add an error css class to input fields

tomfarm:
This commit adds the "field_with_errors" css class
to input fields if there is a matching error on form.errors.
This makes it easier to style the errors on a changeset form
(ecto…) without adding this manually for every input in the form.

Hi!

I'm not sure if this is a fit for the phoenix_html project or if this has been discussed already.
Also note that I'm just starting to get into Elixir so code might not be idiomatic Elixir.


josevalim:
I don't want to make this part of the input. It is coupling concerns that should not be coupled. We will have more news on this topic soon. :) Thanks for the PR!

tomfarm:
ok :)

PR #31, opened 2015-09-13
Nested & attr-less content_tag throws exception

mgwidmann:
Found this while trying to build and unordered list here: https://github.com/mgwidmann/scrivener_html/blob/master/lib/scrivener/html.ex#L112

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -48,8 +48,12 @@ defmodule Phoenix.HTML.Tag do
       ...> end
       {:safe, ["<p class=\"test\">", "Hello", "</p>"]}
   """
-  def content_tag(name, content) when is_atom(name) do
+  def content_tag(name, content) when is_atom(name) and not is_list(content) do
     content_tag(name, content, [])
+  end
+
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/tag.ex
```diff
@@ -48,10 +48,14 @@ defmodule Phoenix.HTML.Tag do
       ...> end
       {:safe, ["<p class=\"test\">", "Hello", "</p>"]}
   """
-  def content_tag(name, content) when is_atom(name) do
+  def content_tag(name, content) when is_atom(name) and not is_list(content) do
     content_tag(name, content, [])
   end
 
+  def content_tag(name, [do: block]) when is_atom(name) do
```
josevalim: Can you please move this clause up and ensure this still works:

    content_tag :p, ["hello", ?\s, "world"]

Thank you!

mgwidmann: Sure I'll add a test for it in a little bit.

mgwidmann:
@josevalim all good to go.

I noticed theres no integrated test setup... But I checked and all the tests pass on the last commit. :+1:

PR #35, opened 2015-09-23
link("here", to: nil) says :to was not provided

mgwidmann:
Its very confusing when the `:to` option was provided but it says that it was not provided (but that actually the code cannot tell that the value was `nil`)

This improves the error message and returns the options to the user so that they can decipher for themselves what went wrong. In the case where nil was provided, it will be more obvious.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -72,7 +72,7 @@ defmodule Phoenix.HTML.Link do
     {method, opts} = Keyword.pop(opts, :method, :get)
 
     unless to do
-      raise ArgumentError, "option :to is required in link/2"
+      raise ArgumentError, "expected non-nil value for :to in link/2, got: #{inspect to}"
     end
 
     if method == :get do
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/link.ex
```diff
@@ -67,12 +67,12 @@ defmodule Phoenix.HTML.Link do
...
 
-  def link(text, opts) do
-    {to, opts} = Keyword.pop(opts, :to)
+  def link(text, original_opts) do
+    {to, opts} = Keyword.pop(original_opts, :to)
     {method, opts} = Keyword.pop(opts, :method, :get)
 
     unless to do
-      raise ArgumentError, "option :to is required in link/2"
+      raise ArgumentError, "option :to is required in link/2 and its value cannot be nil, received: #{inspect original_opts}"
```
josevalim: I don't think we need `original_opts`, this should be enough:

    raise ArgumentError, "expected non-nil value for :to in link/2, got: #{inspect to}"

Can you please amend? Thank you!

mgwidmann:
@josevalim changed as you suggested, should I rebase or is it ok?

josevalim:
Please squash. It should be one commit. :)

-- 


mgwidmann:
@josevalim :+1:

josevalim:

PR #38, opened 2015-10-03
Adds escape_javascript method

JustMikey:
The regex for the method is taken from RoR's `javascript_method` but with 2 elements removed from it, specifically `\342\200\250` and `\342\200\251` because I am simply not sure what those are for and I can't implement them anyway so it's open for an enhancement.

Also the list of matches is implemented as a module attribute, which may be wrong, and if it is, then I guess it won't be problematic to make it a private method or whatever.

If this PR is just completely wrong then I hope that this method will still be implemented in one form or another but by somebody more experienced

### DIFF
--- a/lib/phoenix_html.ex
+++ b/lib/phoenix_html.ex
@@ -158,17 +158,8 @@ defmodule Phoenix.HTML do
     IO.iodata_to_binary(iodata)
   end
 
-  @escape_javascript_map %{
-    "\\"    => "\\\\",
-    "</"    => "<\/",
-    "\r\n"  => "\n",
-    "\r"    => "\n",
-    "\""    => "\\\"",
-    "'"     => "\\'"
-  }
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @JustMikey! The fastest way to implement it would be to not use regex and rather something like follows:

https://github.com/elixir-lang/plug/blob/master/lib/plug/html.ex#L15-L33

JustMikey:
Oh I see, thank you. I will try to rewrite it in the following way :)

JustMikey:
I pushed it to the github to test it but it automatically added new changes here and unfortunally I didn't know it would do it so it was not intended. This new code doesn't seem to work properly so I apologize for all this, won't happen again.

josevalim:
No worries, thanks for the contributions  @JustMikey !

PR #39, opened 2015-10-04
Adds escape_for_javascript method

JustMikey:


lib/phoenix_html.ex
```diff
@@ -157,4 +157,37 @@ defmodule Phoenix.HTML do
...
+
+  @escapes_for_javascript [
+    {?", "\\\""},
+    {?', "\\\'"},
+    {?\\, "\\\\"},
+    {?\n, "\\n"}   
+  ]
+
+  Enum.each @escapes_for_javascript, fn { match, insert } ->
+    defp escape_char(unquote(match)), do: unquote(insert)
```
chrismccord: let's namespace this as `escape_js_char` or similar

lib/phoenix_html.ex
```diff
@@ -157,4 +157,37 @@ defmodule Phoenix.HTML do
   def safe_to_string({:safe, iodata}) do
     IO.iodata_to_binary(iodata)
   end
+
+  @doc """
+  Escapes quotes (double and single) and double backslashes
+
+  This method is extremely helpful in JavaScript responses when there is a need
+  to escape html rendered from other templates, like in the following:
```
chrismccord: "This method is useful in ..." :)

chrismccord:
I have left just a couple small comments. Please squash your commits and I'll need @josevalim to signoff, but looks good from here. Thanks!

lib/phoenix_html.ex
josevalim: This *function*. :)

chrismccord: 😳

JustMikey:
Squashed the commits and added the changes

chrismccord:
<3 <3 <3

JustMikey:
Maybe it's worth bumping the version a little?

josevalim:
I have other things to merge than I will bump it. :)

josevalim:
@JustMikey which reference did you use to pick those values? Rails escapes many other characters: https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/javascript_helper.rb#L6-L17

JustMikey:
@josevalim ok :) Well, I played with it on practice and kind of added the bare minimum so it's probably up for improving.

Issue #36, opened 2015-09-27
content_tag doesn't render integers as content


JustMikey:
If used in console, `content_tag(:div, 23, class: "some-class")` generates proper html, but when it's used in a template, inside `<%= %>`, it renders only the `div` tag but it's empty (no 23 in it)

If there is no way to change it, it would be great to mention about not using integers as content somewhere

PR #40, opened 2015-10-06
Improved tag.ex docs

JustMikey:
Added a comment regarding passing numbers as content to `content_tag`.


josevalim:
Integers should actually be fine because they implemented the Phoenix.HTML.Safe protocol. :)

JustMikey:
@josevalim hmm, could you please check this issue https://github.com/phoenixframework/phoenix_html/issues/36 then? Maybe I just don't undeerstand something

PR #42, opened 2015-10-10
Adds a function to create a button type element in a form.

felipesere:
This PR adds a function to create a `<button>Foo</button>` element to a form.
Using </button> is more semantic than using `<input type="submit">`.

One thing I could add is a constraint on the `:type` attribute as only four values are allowed:
`:button`, `:reset`, `submit` and `:menu`. Not putting the attribute in is equivalent to `type="submit"`.

See https://html.spec.whatwg.org/multipage/forms.html#the-button-element for more details.

### DIFF
--- a/test/phoenix_html/form_test.exs
+++ b/test/phoenix_html/form_test.exs
@@ -320,6 +320,19 @@ defmodule Phoenix.HTML.FormTest do
           ~s(<input class="btn" type="submit" value="Submit">)
   end
 
+  ## button/2
+  test "button/2" do
+    assert safe_to_string(button("Press")) ==
+           ~s(<button>Press</button>)
+
+    assert safe_to_string(button("Press", class: "btn")) ==
+           ~s(<button class="btn">Press</button>)
+
+    assert safe_to_string(button("Press", [class: "btn", type: :submit])) ==
+           ~s(<button class="btn" type="submit">Press</button>)
+
+  end
+
   ## reset/2
 
   test "reset/2" do
--- a/test/phoenix_html/form_test.exs
+++ b/test/phoenix_html/form_test.exs
@@ -320,19 +320,6 @@ defmodule Phoenix.HTML.FormTest do
           ~s(<input class="btn" type="submit" value="Submit">)
   end
 
-  ## button/2
-  test "button/2" do
-    assert safe_to_string(button("Press")) ==
-           ~s(<button>Press</button>)
-
-    assert safe_to_string(button("Press", class: "btn")) ==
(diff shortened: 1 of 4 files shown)


felipesere:
@chrismccord @josevalim Can you guys maybe have a quick peek at this PR?

test/phoenix_html/form_test.exs
```diff
@@ -320,6 +320,19 @@ defmodule Phoenix.HTML.FormTest do
...
 
+  ## button/2
+  test "button/2" do
+    assert safe_to_string(button("Press")) ==
+           ~s(<button>Press</button>)
+
+    assert safe_to_string(button("Press", class: "btn")) ==
+           ~s(<button class="btn">Press</button>)
+
+    assert sage_to_string(button("Press", [class: "btn", type: :submit])) ==
```
josevalim: typo

josevalim:
Btw, we already have a button function here https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/link.ex#L130-L148 and they will certainly clash. Given that this one is really a wrapper for content_tag, I would rather advise to just use HTML tags or content_tag... thoughts?

felipesere:
Maybe I can rename my function to `form_button`?
I'd really like to write HTML5 without having to write content_tag everywhere :smile:

josevalim:
`form_button` doesn't help much. Honestly there other one is more like form_button but we can't change it. :( Another option is button_tag but I am not a big fan as well.

felipesere:
How about just extending the   `button(...)` works? `button("some value")` would generate `<button>some value</button>` while the old behaviour will remain unchanged.

It may be kind of messy, especially being in the `link` file...

josevalim:
Yes, it would be confusing. :( I think direct html or content tag will be the way to go for this one.

felipesere:
I wanted to see how confusing the code would so I gave it a quick stab.
@josevalim do you mind taking another look? 
It's still backwards compatible, but makes `:to` an optional parameter rather than a mandatory one.
I'll extend the doc accordingly if you think is is anywhere usable...

josevalim:
Yeah, I am :-1: because it behaves completely different depending on the arguments. One has its own form, the other does not. :( Thank you for giving it a shot anyway!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

felipesere:
Its ok :smile_cat:  
It was a fun experiment, let's see if I can spot some other issues around elixir/phoenix to tackle!
:heart: :blue_heart: :green_heart:

PR #43, opened 2015-10-11
Minor additions to the docs for `Phoenix.HTML.Form`

jeffkreeftmeijer:
1. "three distinct scenario" => "three distinct scenarios"
2. Eplain that inputs will be set to values from the request params for
   connection data forms
3. Explain what the `:enforce_utf8` option does when it's not disabled.


josevalim:

Issue #32, opened 2015-09-20
Pass ecto models to multiple_select


NilsLattek:
I am currently playing around with the phoenix framework and I like the similarities to rails!
One thing that I think is still missing is the proper handling of many to many relationships. Regarding the changes required in the ecto framework there is already a ticket: https://github.com/elixir-lang/ecto/issues/183

I think another change would be to allow the multiple_select helper to work with ecto models. So that we can do something like this:
```elixir
# in the controller:
# all_users = Repo.all(User)
<%= multiple_select f, :users, @all_users %>
```

And the form helper could fetch all currently assigned users using the ```value_from(form, field)``` method and set the "selected" property on the option elements. In Rails there is the http://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-collection_check_boxes to handle this case, but I think it is much nicer to use a multiple select so you can do autocomplete input fields.

josevalim:
Just filter the data. In your controller:

    users_select = Repo.all from u in User, select: {u.name, u.id}

There is no way Ecto nor Phoenix can guess which fields we should use. Better to be explicit. You can also use a Enum.map later:

    <%= multiple_select f, :users, Enum.map(@all_users, &{&1.name, &1.id}) %>

NilsLattek:
Thanks for your quick reply. The other form helpers like text_input set their value automatically from the form model. With multiple_select I have to pass in the current values manually (and transform them again) else the options won't be selected:

```elixir
<%= multiple_select f, :users, @all_users, value: Enum.map(@changeset.model.users, fn(u) -> u.id end) %>
```
But when the form is rendered because of an error, the value property should be filled in with the current ```changeset.params.users``` attribute not ```changeset.model.users```. At least this is the way the text_input helper works. This is handled in the private ```value_from``` method.
It looks like a lot of duplicate code if I do this in all forms manually.

In Rails the helper assumes some default names for the fields, but it is possible to override these:
```ruby
collection_check_boxes(:post, :author_ids, Author.all, :id, :name_with_initial)
```

josevalim:
> But when the form is rendered because of an error, the value property should be filled in with the current changeset.params.users attribute not changeset.model.users. At least this is the way the text_input helper works. This is handled in the private value_from method.

I see the issue now. The issue is that we are not correctly marking the proper fields as chosen when the data comes through the form.

[...]

PR #45, opened 2015-10-21
Should preserve the field value in multiple_select

bryansray:
This allows the multiple_select/4 function to preserve the values that were submitted to the model.

This resolves issue #32


josevalim:

PR #46, opened 2015-10-21
Adds number helper module

JustMikey:
This module adds two functions to simplify some operations with numbers. I wanted to add some other but they either required some i18n mechanism or were too general but I can add something else if you have any offers. 

Also the code is probably lousy and if it is so, what should I change?

### DIFF
--- a/lib/phoenix_html/number_helper.ex
+++ b/lib/phoenix_html/number_helper.ex
@@ -1,0 +1,138 @@
+defmodule Phoenix.HTML.NumberHelper do
+  @moduledoc """
+  Provides helper functions for number transformations
+  """
+
+  @default_delimiter ","
+  @default_separator "."
+  @default_precision 3
+
+  @doc """
+  Formats a number grouping thousands using the delimiter.
+
+  Function accepts following options:
+
+    * `:delimiter` - The thousands delimiter (defaults to “,”).
+
+    * `:separator` - The separator between the integer and fractional digits (defaults to “.”).
+
+  Examples:
+
+      iex> number_with_delimiter(12345)
+      "12,345"
+      iex> number_with_delimiter(12345, delimiter: " ")
+      "12 345"
+      iex> number_with_delimiter(12345.678, delimiter: " ", separator: ",")
+      "12 345,678"
+  """
+  @spec number_with_delimiter(number, Keyword.t) :: binary
+  def number_with_delimiter(number, opts \\ [])
+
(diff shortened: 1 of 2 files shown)


chrismccord:
So I am unsure how far to go with helper functions. Since we have this wonderful functional language, a separate package should be able to be dropped in and functions imported. So for me we either need to go all in on adding generalized formatting helpers or defer to another package. If we just add a few, they aren't really gaining us or users much, so I'm not sure yet which way to go. I worry about things becoming a grab bag of functions over time // @josevalim ?

JustMikey:
I think it may make sense to add some general commonly-used functions here and move the more specific functions, if there is a need for those, somewhere else. But I don't know, it's up to you.

bryansray:
I would definitely like this to see these as a separate package at a minimum. I could also use a CurrencyHelper at some point as well. Thanks for the code @JustMikey

JustMikey:
So what shall we do? // @josevalim

josevalim:
Sorry, I won't have an answer soon. Please continue using it in your projects as a separate thing until I solve everything from my side and can focus on Phoenix.HTML again.

josevalim:
And thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

paulcsmith:
FWIW I like the idea of having commonly used functions in one place so you don't have to search all over for functions that are often used. However I can see the case for splitting it into another module.

Either way though, if this is added, I think it should be added to Phoenix since these things could be useful outside of generating HTML. Could be used for APIs as well

szymon-jez:
Actually there is now a decent library that offers number formatting functions https://github.com/danielberkompas/number

BTW What I'm missing is a standard way for number localization in Phoenix/Elixir. I had to implement it myself (helping myself with some (copy-pasted) code from the above mentioned library).
Coming from Rails I was surprised that Phoenix itself (which is OK) or it's tooling and docu (which is IMHO not OK) missed covering this very common topic and pointing to some solutions.

chrismccord:
> or it's tooling and docu (which is IMHO not OK) missed covering this very common topic and pointing to some solutions.

We are limited on guide contributors, so if you're interested in improving this area, we'd love a guide contribution on this topic. Thanks!

josevalim:
Yes, I agree with @chrismccord here. I think separate packages makes more sense, because nothing in this is really specific to phoenix_html.

josevalim:
Thanks everyone!

PR #47, opened 2015-11-08
Try to hint why an expected view assign is missing

bklang:
This is an alternative approach to my previous pull request, phoenixframework/phoenix#1342.  I'm submitting this because I found the framework behavior surprising and it took me a while to figure it out. I understand the motivation for not accepting my previous PR, so I'm hopeful that this log message will save time for others who encounter a similar issue.

I'm not 100% confident in the terminology I chose (parent vs. child view template), so please feel free to offer alternative wording suggestions.

### DIFF
--- a/lib/phoenix_html/engine.ex
+++ b/lib/phoenix_html/engine.ex
@@ -83,7 +83,13 @@ defmodule Phoenix.HTML.Engine do
     case Dict.fetch(assigns, key) do
       :error ->
         raise ArgumentError, message: """
-        assign @#{key} not available in eex template. Available assigns: #{inspect Dict.keys(assigns)}
+        assign @#{key} not available in eex template.
+
+        Please make sure all proper assigns have been set. If this
+        is a child template, ensure assigns are given explicitly by
+        the parent template as they are not automatically forwarded.
+
+        Available assigns: #{inspect Dict.keys(assigns)}
         """
       {:ok, val} -> val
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you! What if we say something like: "please make sure all expected assigns have been given in the render call". Would that be enough in your case?

bklang:
I don't think that would have been enough of a hint for me, because I was thinking that the only render call that mattered was the one in the controller. The thing that cost me time was realizing that there was a second render call in the containing view where I needed to repeat the assigns.

bklang:
Also, I see the spec failure and will fix it when we settle on wording.

josevalim:
Sorry for the delay, what about this:

```elixir
"assign @#{key} not available in eex template.\n\n" <>

"Please make sure all proper assigns have been set. If this " <>
"is a child template, ensure assigns are given explicitly by " <>
"the parent template as they are not automatically forwarded.\n\n" <>

"Available assigns: #{inspect Dict.keys(assigns)}"
```

chrismccord:
:+1: to @josevalim's tweaks

bklang:
Same, :+1: from me. Thank you for the feedback. I'll update the PR and the tests.

bklang:
@josevalim @chrismccord all set :)

chrismccord:
<3 <3 <3

PR #48, opened 2015-11-17
Add Phoenix.HTML.Format.text_to_html/2

michalmuskala:
Because I recently needed this functionality I implemented it in a package [simple_format](https://github.com/michalmuskala/simple_format), but @josevalim advised adding it directly to `Phoenix.HTML`. So here it is :smiley:

### DIFF
--- a/lib/phoenix_html.ex
+++ b/lib/phoenix_html.ex
@@ -12,6 +12,8 @@ defmodule Phoenix.HTML do
 
     * `Phoenix.HTML.Link` - functions for generating links and urls;
 
+    * `Phoenix.HTML.Format` - functions for formatting text;
+
   ## HTML Safe
 
   One of the main responsibilities of this module is to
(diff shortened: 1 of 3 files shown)


michalmuskala:
I also considered adding it as `Phoenix.HTML.Format.simple/2` instead. Would that be better?

josevalim:
Thank you! I was wondering about the name as well. `simple/2` is too unclear though. I would go with `simple_format/2` or something like `text_to_html/2`. :+1: for `Phoenix.HTML.Format` as module name.

chrismccord:
`text_to_html` is clearer to me than `simple_format`, or maybe `newlines_to_html`

michalmuskala:
Updated. I've decided to go with `Format.text_to_html/2`.

josevalim:

lib/phoenix_html.ex
```diff
@@ -12,6 +12,8 @@ defmodule Phoenix.HTML do
 
     * `Phoenix.HTML.Link` - functions for generating links and urls;
 
+    * `Phoenix.HTML.Format` - functions for formatting text;
```
josevalim: Oops, please update the `__using__` callback so we also import `Format` by default.

michalmuskala: Done: https://github.com/phoenixframework/phoenix_html/pull/49

PR #50, opened 2015-12-08
Fix for cast example in Phoenix.HTML.Form docs

bhserna:
In the section of "With model data" for the `Phoenix.HTML.Form` docs, in the definition of the `User.changeset/2` function, the example calls a `cast/2` function but I think that the intention here is to call the `Ecto.Changeset.cast/4` function. Is that is the case this change could help.


chrismccord:
Thanks!

PR #57, opened 2016-01-04
Add support for tags with lists as attribute values

vorce:
Handy when you have arrays in an ecto model for example.

Not sure if this is the right place to do this or if it's even desirable.


josevalim:
Thank you @vorce! I am :-1: because different attributes will likely use different formatting for lists. For example, for CSS classes we would just use Enum.join, without comma. So it is probably better to be explicit.

vorce:
Ah yes - agreed! Thank you for the feedback.

PR #58, opened 2016-01-07
More helpful error when missing translate_error

henrik:
Since this is very specific, I don't have great hopes that it'll be merged as-is.

But I've seen people on Slack that were bit by this, three times in the last few days, and I haven't even been looking. So I'm thinking a small, simple fix that we can remove by Phoenix 2 or earlier can be a cheap way of avoiding some upgrade pain.

### DIFF
--- a/lib/phoenix_html/safe.ex
+++ b/lib/phoenix_html/safe.ex
@@ -70,4 +70,9 @@ end
 
 defimpl Phoenix.HTML.Safe, for: Tuple do
   def to_iodata({:safe, data}), do: data
+  def to_iodata(value) do
+    raise Protocol.UndefinedError,
+      protocol: @protocol,
+      value: value
+  end
 end
--- a/test/phoenix_html/safe_test.exs
+++ b/test/phoenix_html/safe_test.exs
@@ -22,4 +22,10 @@ defmodule Phoenix.HTML.SafeTest do
     assert Safe.to_iodata(1.0) == "1.0"
     assert Safe.to_iodata({:safe, "<foo>"}) == "<foo>"
   end
+
+  test "Phoenix.HTML.Safe given an invalid tuple" do
+    assert_raise Protocol.UndefinedError, fn ->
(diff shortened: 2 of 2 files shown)


lib/phoenix_html/safe.ex
```diff
@@ -70,4 +70,11 @@ end
 
 defimpl Phoenix.HTML.Safe, for: Tuple do
   def to_iodata({:safe, data}), do: data
```
josevalim: Good catch. I believe we should raise a Protocol.UndefinedError, here is an example from Elixir:

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/list/chars.ex#L28-L31

That would at least improve the error message.

I am not sure if we should do the translate_error bit though. Theoretically users should not run into this because we have released a new major phoenix_ecto version, so they would just run into this if updating phoenix_ecto explicitly.

henrik: I think one reason (maybe the only?) people run into this is that they have the "old" Phoenix book that describes the old way of doing things, but they're generating a new 1.1 app.

Maybe that's going away when the book is updated (if it hasn't already), I dunno.

I feel that this change is small and isolated enough that I think the cost for being possibly-too-helpful is low, but I'm not a maintainer, so it's definitely not my call :)

Let me know where you'd like this to go and I'll make the change. Or feel free to do it yourself outside this PR if you have a specific idea and want to run with it :)

josevalim: > Maybe that's going away when the book is updated (if it hasn't already), I dunno.

It will be this week.

> Let me know where you'd like this to go and I'll make the change. 

Would you send a change that at least raises Protocol.UndefinedError for now without any extra description? Then it is just a matter of adding the description if the problem persists. :)

henrik: @josevalim Done! This makes the error more specific at least – not sure if it helps much if you make that mistake, but as you say, it makes it easier to just add a description if needed :) I'll keep an eye out.

josevalim:

PR #59, opened 2016-01-07
Extract Safe to own tests

henrik:
Wasn't where I expected it when I wanted to modify that module.

Also taking the opportunity to make the remaining test names more consistent.


josevalim:

PR #62, opened 2016-01-19
Enhance documentation for data-confirm attribute

sorentwo:
The data-confirm attribute is supported by phoenix_html.js, but instructions for generating the confirm attribute was lacking. This adds an example and adds more explicit direction for adding data attributes.

Additionally, it seems fitting to me that passing a bare `confirm` keyword would automatically generate the `data-confirm` attribute. I originally intended to add support for it, but decided on documenting the current situation instead. If supporting `confirm` directly is preferred/desired, I can add that as well and update the PR.


josevalim:

PR #63, opened 2016-01-24
Form value/id/name helpers were made public.

### LINKED ISSUE #44: Add value_from to the public form API
Every time I need to do something custom I had to use undocumented ways of getting the value out of the form parameter. It feels especially scary when working with nested forms. BTW exposing id and name will be also helpful.

sumerman:
Resolves #44


josevalim:

PR #73, opened 2016-03-17
Fix #64, Incorrect dialyzer type mismatch

### LINKED ISSUE #64: Incorrect dialyzer type mismatch
A newly installed phoenix 1.1.4 app has a problem with html.eex functions

```
index.html.eex:1: The pattern {'safe', _@2} can never match the type binary()
```

I've created a [phoenix app](https://github.com/paulcsmith/dialyze-demo) that you can clone and run `mix do compile, dialyzer.plt, dialyzer` and see the output. This is related to https://github.com/elixir-lang/elixir/issues/4240#issuecomment-176878659 which contains a bit more information about the problem for those that are curious :)

I would open a PR, but I'm still not entirely sure how to mark a clause as generated :S I'd be happy to help if you could provide some guidance, but I'm also ok if you it's easier to just fix by yourself. LMK if you need any other info!

ciastek:
Trying to fix #64 , but have no idea what am I doing.

### DIFF
--- a/lib/phoenix_html/engine.ex
+++ b/lib/phoenix_html/engine.ex
@@ -61,19 +61,21 @@ defmodule Phoenix.HTML.Engine do
     quote line: line, do: Phoenix.HTML.Safe.to_iodata(unquote(literal))
   end
 
-  # We need to check at runtime and we do so by
-  # optimizing common cases.
+  # We need to check at runtime
   defp to_safe(expr, line) do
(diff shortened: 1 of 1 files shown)


josevalim:
Well, the goal of this code is to inline code and by moving it at compile time we are not doing any of the inlining anymore.

josevalim:
And we already do compile time inlining, so we shuold really need runtime inlining there.

PR #74, opened 2016-03-30
Fix form tag example code

tomekowal:
Example usage of `form_tag` had "get" in arguments, but "post" in output.


josevalim:

PR #76, opened 2016-04-17
Fix docs: escape interpolation

petekinnecom:
Fixes the formatting seen here: https://hexdocs.pm/phoenix_html/Phoenix.HTML.html#macros


chrismccord:
thanks!

Issue #75, opened 2016-04-05
Use event delegation in phoenix_html.js


maufl:
Currently, one event handler is attached to each element that uses `data-submit=parent`.
The downside is that  the elements must be present in the DOM when the file is loaded. I think that also means that the script tag with the code **must** come after the elements.

Instead we could use [event delegation](https://davidwalsh.name/event-delegate). The event handler can then be attached to the body and will work for any element, whenever it is loaded in the DOM (via Ajax for example).

chrismccord:
Would you like to send a PR?

maufl:
Yes, I'll probably do that soon. I don't have time today and the issue is not that pressing, I just wanted it have it documented here.

josevalim:
Beautiful, thank you!

PR #77, opened 2016-04-18
Use event delegation in phoenix_html.js #75

Troush:
Refers to #75

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -1,19 +1,12 @@
 'use strict';
 
-// Although ^=parent is not technically correct,
-// we need to use it in order to get IE8 support.
-var elements = document.querySelectorAll('[data-submit^=parent]');
-var len = elements.length;
-
-for (var i = 0; i < len; ++i) {
-  elements[i].addEventListener('click', function (event) {
-    var message = this.getAttribute("data-confirm");
+window.addEventListener('click', function (event) {
+  if(event.target && event.target.matches('a[data-submit=parent]')) {
+    var message = event.target.getAttribute('data-confirm');
     if (message === null || confirm(message)) {
-      this.parentNode.submit();
+      event.target.parentNode.submit();
     };
     event.preventDefault();
     return false;
-  }, false);
-}
-
-;
+  }
+}, false);


josevalim:
Thank you @Troush! Maybe we should drop support for IE8? Folks can write their own tiny phoenix-html script if they need to support older browsers.

priv/static/phoenix_html.js
```diff
@@ -1,19 +1,22 @@
...
+function matches(elm, selector) {
+    var matches = (elm.document || elm.ownerDocument).querySelectorAll(selector),
+        i = matches.length;
+    while (--i >= 0 && matches.item(i) !== elm) {}
+    return i > -1;
 }
 
-;
\ No newline at end of file
+window.addEventListener('click', function (event) {
```
chrismccord: two space indentation please

priv/static/phoenix_html.js
```diff
@@ -1,19 +1,22 @@
...
+	// Although ^=parent is not technically correct,
+	// we need to use it in order to get IE8 support.
+	if(event.target && matches(event.target, 'a[data-submit^=parent]')) {
+	    var message = event.target.getAttribute("data-confirm");
```
chrismccord: use consistent quoting, either use is fine

Troush:
Updated. Without IE8 support, and indentation fixes.
Sorry for indentation :\

chrismccord:
<3 <3 <3

maxgronlund:
Worked great for a short while. how it's broken, keep it up

josevalim:
@maxgronlund I think there is a typo in your message and that made me confused. :) It is working now or is it broken? :)

Troush:
@maxgronlund Describe more detail whats working wrong?

I will happy to fix if something wrong :)

jsteiner:
It seems this change breaks compatibility with `phantomjs` (`2.1.1`, the latest), as `phantomjs` doesn't support `Element.matches`. Adding a polyfill fixed it locally.

Would a fine solution here be to update`phoenix_html` to include the polyfill?

josevalim:
@jsteiner instead of introducing matches, what if we explicitly check for the tag and the attribute of the element since that's the only information we care about? I think it is going to be simpler/faster than including the polyfill? Could you please send a PR?

PR #78, opened 2016-04-22
Add :insert_brs option to text_to_html/2

ericmj:


ericmj:
This gives us a more markdown-like paragraph handling.

PR #79, opened 2016-05-04
Do not add a class to buttons by default

drapergeek:
It seems like an odd decision to add a class to any markup generated without the user opting into that decision. Happy to hear opinions on this decision but I thought I would get the ball rolling.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -112,7 +112,7 @@ defmodule Phoenix.HTML.Link do
   ## Examples
 
       button("hello", to: "/world")
-      #=> <form action="/world" class="button" method="post">
+      #=> <form action="/world" method="post">
             <input name="_csrf_token" value=""><input type="submit" value="hello">
           </form>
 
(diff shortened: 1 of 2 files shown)


josevalim:
We need to otherwise it is impossible to style those forms which is something we do in the default .css that ships as part of Phoenix. We are glad to make it customizable but we can't remove them.

drapergeek:
>We need to otherwise it is impossible to style those forms which is something we do in the default .css that ships as part of Phoenix. We are glad to make it customizable but we can't remove them.

It's already customizable, which I appreciate.

Why should the framework have any opinion on styling by default? That doesn't seem like something that the framework should be in charge of.

josevalim:
This is just one library, it is not like the framework is imposing this choice on you. :) At the time it looked like a reasonable default as otherwise it would be impossible to style them. In any case, we can't simply remove it now without breaking all applications that are relying on it there.

chrismccord:
I agree it's a tradeoff, but you can easily wrap the the link and button functions `import ... except: [link: 2, button: 2]` and use your own behaviour/styling.

drapergeek:
Fair enough, thanks for the awesome library!

PR #80, opened 2016-05-05
Added new form element generator for helper text

arpieb:
I was building out a Phoenix site recently, and coming from a CMS background noticed that on nearly every form element I found myself adding helper text boilerplate like:

`<small class="text-muted">A cron-formatted schedule string</small>`

What I really wanted was something that could generate the boilerplate for me:

`<%= help_text "A cron-formatted schedule string", class: "text-muted" %>`

This PR contains the proposed **help_text/2** functions.  Modeled after the **label/2** and **label/4** functions, it also introduces a do-block variant that would allow for code-based generation of the help text from form data.  I feel that along with all the other helper functions provided to generate consistent form markup, this would be a nice addition.

Thanks for looking!

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1141,4 +1141,47 @@ defmodule Phoenix.HTML.Form do
   def field_name(name, field) when is_atom(name),
     do: "#{name}[#{field}]"
 
+  @doc """
+  Generates standard markup for field help text.
+
+  The form should either be a `Phoenix.HTML.Form` emitted
+  by `form_for` or an atom.
+
+  All given options are forwarded to the underlying tag.
+
+  ## Examples
+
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks! What you wanted instead of a new feature is exactly to define those same functions in your project. You can define your `help_text` functions in your project and import them into your view(s). This solves your boilerplate issue and you can implement it in a way that works for your application, such as the <small> tag in this PR.

arpieb:
Gotcha, thanks for the review!  I was thinking of it being a more general-purpose utility since with a11y you almost never run into a form field these days without a bit of associated helper text, and was looking for a consistently-formatted helper function.

If it doesn't fit in with the Phoenix framework, totally understand.  Maybe next time!  :)

PR #82, opened 2016-05-24
Add `grouped_select` form helper.

ciaran:
Adds a `grouped_select` form helper, which takes a map of group labels => options lists, and creates a select with `<optgroup>` tags.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -657,6 +657,11 @@ defmodule Phoenix.HTML.Form do
   (like a regular list or a map) or any Enumerable where the element
   will be used both as key and value for the generated select.
 
+  ## Optgroups
+
+  When a two-item tuple contains a list as its value,
+  the options will be wrapped in an `<optroup>` using the key as its label.
+
   ## Examples
 
       # Assuming form contains a User schema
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -707,10 +707,41 @@ defmodule Phoenix.HTML.Form do
...
+          </optgroup>
+          ...
+          </select>
+
+  See `select/4` for more information.
+  """
+  def grouped_select(form, field, optgroups, opts \\ []) do
+    opts =
+      opts
+      |> Keyword.put_new(:builder, &grouped_options_for_select/3)
```
chrismccord: `opts = Keyword.put_new(opts, :builder, &grouped_options_for_select/3)`

lib/phoenix_html/form.ex
```diff
@@ -680,6 +683,16 @@ defmodule Phoenix.HTML.Form do
           <option value="user">User</option>
           </select>
 
+      select(form, :country, %{"Europe" => ["UK", "Sweden", "France"], …})
```
josevalim: Can we also support: `%{"Europe" => [{"UK", "uk}, ...]}` and `%{"Europe" => %{"UK" => "uk, ...}}`?

chrismccord: 👍

smeevil:
Thanks for this :) I was on the verge of creating this as well :D

smeevil:
It seems the suggested syntax was already possible, but the restriction of only allowing a list was blocking it. Adding maps to the guard as well seems to work. Please see https://github.com/phoenixframework/phoenix_html/pull/108

PR #83, opened 2016-05-25
Change arg name and documentation for field_value

krummi:
I was looking at the documentation for `field_value` yesterday and was pretty puzzled by this optional `selected` argument for two reasons:

1. The documentation referred both to the `selected` and the `default` argument.
2. The `field_value` is used when fetching values for all inputs, not only for select boxes.

So I decided to make this change. Hopefully it makes sense. ;)


josevalim:

PR #95, opened 2016-07-06
render button tag for submit instead of input

allyraza:
Right now it renders a input tag for submit I think it should be a button, buttons are much easier to style


josevalim:

PR #99, opened 2016-07-12
Fix submit links in phantomjs

jsteiner:
The JavaScript was changed in https://github.com/phoenixframework/phoenix_html/pull/77 to use `Element.matches` to find the correct element. Unfortunately, phantomjs doesn't provide that function.

This fixes that by matching the element manually.

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -8,7 +8,7 @@ function isLinkToSubmitParent(element) {
 }
 
 function didHandleSubmitLinkClick(element) {
-  while (element) {
+  while (element && element.getAttribute) {
     if (isLinkToSubmitParent(element)) {
       var message = element.getAttribute('data-confirm');
       if (message === null || confirm(message)) {
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -1,8 +1,15 @@
 'use strict';
 
-function didHandleSubmitLinkClick(element){
-  while(element) {
-    if(element.matches && element.matches('a[data-submit=parent]')){
+function isLinkToSubmitParent(element) {
+  var isLinkTag = element.tagName === 'A';
+  var shouldSubmitParent = element.getAttribute('data-submit') === 'parent';
+
+  return isLinkTag && shouldSubmitParent;
+}
+
+function didHandleSubmitLinkClick(element) {
+  while (element) {
+    if (isLinkToSubmitParent(element)) {
       var message = element.getAttribute('data-confirm');
       if (message === null || confirm(message)) {
         element.parentNode.submit();
@@ -17,8 +24,9 @@ function didHandleSubmitLinkClick(element){
 
 // for links with HTTP methods other than GET
 window.addEventListener('click', function (event) {
-  if(event.target && didHandleSubmitLinkClick(event.target)) {
+  if (event.target && didHandleSubmitLinkClick(event.target)) {
     event.preventDefault();
     return false;
   }
 }, false);
+
--- a/web/static/js/phoenix_html.js
+++ b/web/static/js/phoenix_html.js
@@ -6,15 +6,15 @@ function isLinkToSubmitParent(element) {
 }
 
 function didHandleSubmitLinkClick(element) {
-  while(element) {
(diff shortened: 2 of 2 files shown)


priv/static/phoenix_html.js
```diff
@@ -1,8 +1,15 @@
 'use strict';
 
-function didHandleSubmitLinkClick(element){
-  while(element) {
-    if(element.matches && element.matches('a[data-submit=parent]')){
+function isLinkToSubmitParent(element) {
+  var isLinkTag = element.tagName === 'A';
+  shouldSubmitParent = element.getAttribute('data-submit') === 'parent';
```
josevalim: Are we missing `var` here?

web/static/js/phoenix_html.js
```diff
@@ -1,6 +1,13 @@
-function didHandleSubmitLinkClick(element){
+function isLinkToSubmitParent(element) {
+  const isLinkTag = element.tagName === 'A';
+        shouldSubmitParent = element.getAttribute('data-submit') === 'parent';
```
josevalim: Let's use `var` instead of `const` and we need to use it on both lines. :) Thank you!

jsteiner: Fixed!

jsteiner:
Actually, hold off on merging this. I just noticed some errors being raised when clicking on other elements. Digging into it now.

jsteiner:
Okay, should be all set :)

web/static/js/phoenix_html.js
```diff
@@ -1,13 +1,20 @@
...
+}
+
+function didHandleSubmitLinkClick(element) {
   while(element) {
-    if(element.matches && element.matches('a[data-submit=parent]')){
+    if(isLinkToSubmitParent(element)) {
       var message = element.getAttribute('data-confirm');
       if (message === null || confirm(message)) {
-        element.parentNode.submit();
+        element.parentElement.submit();
```
josevalim: Do we need this change? I was reading that SVG documents do not have `parentElement`. If we trigger this function for SVG entities, which I think we do, then this fail. Maybe we should stick with `parentNode`?

jsteiner: We need something to prevent this from climbing to the document node, since document isn't an `Element` and doesn't implement `getAttribute`.  Without something to prevent that, it errors any time you click something that isn't an `a[data-confirm=parent]`. Another option would be to set the while loop to `while (element && element.getAttribute)`.

josevalim: I think the `while` loop approach is a good one. Shall we do that then?

jsteiner: Also, it looks like `SVGElement` inherits from `Element`:

> The Element interface represents an object of a Document. This interface describes methods and properties common to all kinds of elements. Specific behaviors are described in interfaces which inherit from Element but add additional functionality. For example, the HTMLElement interface is the base interface for HTML elements, while the SVGElement interface is the basis for all SVG elements.

https://developer.mozilla.org/en-US/docs/Web/API/Element

josevalim: @jsteiner the problem seems to be IE as usual, see second response: http://stackoverflow.com/questions/8685739/difference-between-dom-parentnode-and-parentelement :(

Troush:
We can provide polyfill for `Element.matches`
```
function selectorMatches(el, selector) {
	var p = Element.prototype;
	var f = p.matches || p.webkitMatchesSelector || p.mozMatchesSelector || p.msMatchesSelector || function(s) {
		return [].indexOf.call(document.querySelectorAll(s), this) !== -1;
	};
	return f.call(el, selector);
}
```
That will save a clean code flow and fix phantomjs/IE problem.

josevalim:
@Troush we decided to not use the matches polyfill because we only care about the tag and named attributes. Maybe we should drop our solution in favor of it though.

jsteiner:
How big of a concern is performance? While I prefer the polyfill stylistically, it seems matching directly on the attributes of the element would be more performant than `querySelectorAll` (which I assume traverses the entire DOM). I imagine this wouldn't be an issue for 99% of people, but for framework code it seems performance may be more important here?

PR #107, opened 2016-08-11
Implement remote links.

denispeplin:
I'm playing with Phoenix and trying to implement remote forms and links using [rails-ujs](https://github.com/rails/jquery-ujs). The only discussion I found related this topic is phoenixframework/phoenix#713.
Remote forms aren't require any efforts, it's easy to set `"data-remote": "true"` and plug `rails-ujs`. But remote links aren't implemented in Phoenix.
So I gave them a try.
With `:get` links it was easy, but `:post` and other links, unlike Rails, were implemented in surprisingly different way, via forms, with links for submit.
So for `rails-ujs` to work, I had to put `"data-remote": "true"` inside forms itself, and also to replace those links with inputs with type `submit`.

As for pull request itself, I think I've introduced some more complexity and ugliness to existing code, and I wish to refactor it before merge, if it generally can be accepted.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -74,18 +74,31 @@ defmodule Phoenix.HTML.Link do
   def link(text, opts) do
     {to, opts} = Keyword.pop(opts, :to)
     {method, opts} = Keyword.pop(opts, :method, :get)
+    {remote, opts} = Keyword.pop(opts, :remote, false)
 
     unless to do
       raise ArgumentError, "expected non-nil value for :to in link/2, got: #{inspect to}"
     end
 
     if method == :get do
+      if remote do
+        opts = Keyword.put(opts, :"data-remote", "true")
+      end
       content_tag(:a, text, [href: to] ++ opts)
     else
-      opts = Keyword.put_new(opts, :rel, "nofollow")
-      {form, opts} = form_options(opts, method, "link")
+      {form, opts} = form_options(opts, method, "link", remote)
       form_tag(to, form) do
-        content_tag(:a, text, [href: "#", data: [submit: "parent"]] ++ opts)
+        if remote do
+          opts =
+            opts
+            |> Keyword.put_new(:to, to)
+            |> Keyword.put_new(:type, "submit")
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you, @denispeplin. However, I am slightly confused. Why can't we use the link to submit the remote form as well? This seems to be rather a problem with rails-ujs than Phoenix. No?

denispeplin:
That's right, `rails-ujs` can't send such forms. But forms generally supposed to be sent with inputs type `submit` or buttons, aren't they? Anyway, I'll try to use [vanilla-ujs](https://github.com/hauleth/vanilla-ujs) on forms with links and will tell you about results.

chrismccord:
Let us know how that goes, but we would rather this kind of feature be supported by a third party library that builds on top of phoenix_html than phoenix itself as rails-ujs is rails specific. Thanks!

denispeplin:
I checked `vanilla-ujs`, it is a very light version of `rails-ujs`, which is supposed to work without jquery. Out of box, it expects server to return `text/javascript`, and Phoenix returns `application/javascript`. After changing this expectation, `vanilla-ujs` works well with remote forms, but like `rails-ujs`, only if they contain input type `submit`, not link, as default Phoenix (generated from links) forms do. Also, unlike `rails-ujs`, `vanilla-ujs` have problems remotely submitting `get` links, because it doesn't supply `csrf` for such submits. I didn't fix that because I personally have no need for `vanilla-ujs` yet, I'm using jquery in my projects anyway.

Regarding `rails-ujs`, I can't agree it's Rails specific, I successfully used it in Go project, and it works fine out of the box. Actually, project is called `jquery-ujs`, and only `rails` name in it is top-level object name, other things are not rails-specific: [rails](https://github.com/rails/jquery-ujs/blob/master/src/rails.js). I'm sure that `vanilla-ujs` also can be used after few small modifications.

I think that remote ajax forms, although very simple one, is one of the most powerful 'magic' features of Rails, and very useful one for those who don't build SPA. I was surprised I didn't find this feature already implemented in Phoenix. But I found that responding with javascript is already implemented, so the part that is missing for full support of remote forms is really small.

PR #108, opened 2016-08-13
Support for suggested syntax by @josevalim

smeevil:
It seems to support the two other version that @josevalim suggested did already work. The only thing needed was to easy the guard on line 738 of lib/phoenix_html/form.ex to also allow for maps

I've added two more tests that cover the suggested syntax and merged the current master of phoenixframework/phoenix_html.

I've tested this on one of my own projects and everything seems fine.

Thank you @ciaran for the main work :)

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -657,6 +657,11 @@ defmodule Phoenix.HTML.Form do
   (like a regular list or a map) or any Enumerable where the element
   will be used both as key and value for the generated select.
 
+  ## Optgroups
+
+  When a two-item tuple contains a list as its value,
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -657,6 +657,11 @@ defmodule Phoenix.HTML.Form do
   (like a regular list or a map) or any Enumerable where the element
   will be used both as key and value for the generated select.
 
+  ## Optgroups
+
+  When a two-item tuple contains a list as its value,
```
josevalim: "two-item tuple or a map"?

josevalim:
Just one tiny comment and we are good to go! :D

josevalim:

PR #110, opened 2016-08-19
Add example for a list of ecto schema's to select

barttenbrinke:
I was struggling to find out how to actually map a list of ecto schema's to the select tag method and I found it strange that such a common use case was not mentioned in the docs. So I tried to fix that :)


josevalim:

PR #112, opened 2016-09-16
Allow the submit to take a `do` option

drapergeek:
This utilizes the `content_tag` option for taking a `do` block. The primary use case we're looking to cover is putting an SVG inside the button such as:

```elixir
<%= submit do %>
  <%= content_tag(:svg) %>
<% end %>
```


josevalim:

PR #113, opened 2016-09-16
Use a button instead of an input for the button/2

drapergeek:
This brings the `button` function in line with the change made in #95 and keeps us consistently using the `button` tag instead of the `input` tag. Also the fact that the `button` function doesn't return a button just seems weird 😄 

It also makes it easier for my next proposed change which is going to be allowing a `do:` option on the `button` function.


josevalim:

PR #115, opened 2016-09-20
Support multiple attribute on file_input (#114)

### LINKED ISSUE #114: Document how to receive multiple files with file_input
The docs of `file_input` don't mention anything about `multiple` which led me to thinking that just adding `multiple: true` should do the trick.

Turns out you also have to add `name: "files[]"` or at least something with `[]` at the end to receive more then one file.

It would be great if this would be documented. I don't really get the cause of this so if someone could point that out, this would be great as well.

Fahrradflucht:
Thanks for making me do this @josevalim now I get why `[]` needs to be there.

Closes #114


josevalim:

PR #116, opened 2016-09-21
Submit nearest form even if not direct descendent

troygnichols:
Hi! I'm not positive this is the right thing to do, but it solved a problem for me.

Say I have a template like this:

```
<%= form_for ... do %>
  <div id="foo">
    <%= link to: "/somewhere", class: "fa fa-save", data: [submit: "parent"] %>
  </div>
<% end %>
```

(I use `<a>` instead of `<submit>`  because I need to style the submit button with font-awesome icons e.g.)

With the original code, it tried to submit its immediate parent`div#foo`, which of course doesn't work. This PR looks for the nearest parent that is a form and submits that.

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -7,12 +7,22 @@ function isLinkToSubmitParent(element) {
   return isLinkTag && shouldSubmitParent;
 }
 
+function getClosestForm(element) {
+  while (element && element !== document && element.nodeType === Node.ELEMENT_NODE) {
+    if (element.tagName.toLowerCase() === 'form') {
+      return element;
+    }
+    element = element.parentNode;
+  }
+  return null;
+}
+
 function didHandleSubmitLinkClick(element) {
   while (element && element.getAttribute) {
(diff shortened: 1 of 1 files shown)


josevalim:
@troygnichols unfortunately closest is not supported on internet explorer and on mobiles. :(

troygnichols:
@josevalim Ah, okay. I've updated to use no-frills JS now.

priv/static/phoenix_html.js
```diff
@@ -7,12 +7,22 @@ function isLinkToSubmitParent(element) {
   return isLinkTag && shouldSubmitParent;
 }
 
+function getClosestForm(element) {
+  while (element && element !== document && element.nodeType === Node.ELEMENT_NODE) {
+    if (element.tagName.toLowerCase() === 'form') {
```
josevalim: We should use `element.tagName === 'FORM'` to follow the same pattern we used in the function above.

troygnichols:
That makes sense. Updated.

josevalim:

PR #117, opened 2016-09-29
Add csrf_meta_tag helper

jalkoby:
As Phoenix is js agnostic there is not much js tools outbox as in Rails. So I've made a port of jquery-ujs for phoenix - [phoenix_ujs](https://github.com/jalkoby/phoenix_ujs). 
The only problem is that phoenix_ujs doesn't have csrf meta tag helpers, so I've added it. Plus this helper tell about other "magic" params which phoenix_html expects.

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -7,6 +7,8 @@ defmodule Phoenix.HTML.Tag do
   import Plug.CSRFProtection, only: [get_csrf_token: 0]
 
   @tag_prefixes [:aria, :data]
+  @csrf_param "_csrf_token"
+  @method_param "_method"
 
   @doc ~S"""
   Creates an HTML tag with the given name and options.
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/tag.ex
```diff
@@ -7,6 +7,7 @@ defmodule Phoenix.HTML.Tag do
   import Plug.CSRFProtection, only: [get_csrf_token: 0]
 
   @tag_prefixes [:aria, :data]
+  @csrf [param: "_csrf_token", method: "_method"]
```
josevalim: Please define two separate module attributes. If you do `@csrf[:method]`, it is still injecting the whole list at compile time.

lib/phoenix_html/tag.ex
```diff
@@ -224,9 +225,21 @@ defmodule Phoenix.HTML.Tag do
...
+  @doc """
+  Generates a meta tag with CSRF information.
+
+  ## Tag attributes
+    `content`      - a valid csrf token
+    `csrf-param`   - a request parameter where expected csrf token
+    `method-param` - a request parameter where expected a custom HTTP method
+  """
+  def csrf_meta_tag do
+    raw ~s'<meta charset="UTF-8" name="csrf-token" content="#{get_csrf_token()}" csrf-param="#{@csrf[:param]}" method-param="#{@csrf[:method]}">'
```
josevalim: Please use the proper `tag` to generate this to ensure we won't be vulnerable to xss attacks in the future in case somebody decides add new fields.

lib/phoenix_html/tag.ex
jalkoby: ohh, I didn't think about it. thanks for pointing the problem

jalkoby:
the changes are done

//cc @josevalim

josevalim:

PR #119, opened 2016-10-12
Remove unneeded function head

iamvery:
The last definition of html_safe/1 should cover lists fine as the protocol is implemented for List. Afaik, you generally shouldn't reference the implementation-specific module when using protocols. The only reason I could imagine this might have been intentional is if there's some performance gain from it.


josevalim:
Yes, it was added for optimization purposes. It was added before protocol consolidation but, since it is there, I would still keep it. :)

PR #121, opened 2016-11-09
Allow passing a `do:` option to the button function

drapergeek:
We often wrap a button around svgs for things like deleting a model. This would give us markup something like:

```html
<button class="delete-icon-button">
  <svg>
    <use xlink:href="#icon-cancel-circle">
  </svg>
</button>

```

now this could be used with phoenix:
```html
<%= button to: "/", class: "delete-icon-button" do %>
  <svg>
    <use xlink:href="#icon-cancel-circle">
  </svg>
<% end %>
```

One thing to note: we are passing in `{:safe, do_option}` so that the given option is not HTML escaped. This is a bit of a security issue but I can't imagine a situation where you wouldn't want that functionality. I don't think it makes sense to require `{:safe, "<p>Hi</p>}` inside the `do` option.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -133,7 +133,24 @@ defmodule Phoenix.HTML.Link do
 
   All other options are forwarded to the underlying button input.
   """
+
+  def button(opts, [do: contents]) do
+    {to, form, opts} = extract_button_options(opts)
+
+    form_tag(to, form) do
+      Phoenix.HTML.Form.submit(opts, [do: contents])
+    end
+  end
+
   def button(text, opts) do
+    {to, form, opts} = extract_button_options(opts)
+
+    form_tag(to, form) do
+      Phoenix.HTML.Form.submit(text, opts)
+    end
+  end
+
+  defp extract_button_options(opts) do
     {to, opts} = Keyword.pop(opts, :to)
     {method, opts} = Keyword.pop(opts, :method, :post)
 
@@ -143,9 +160,7 @@ defmodule Phoenix.HTML.Link do
       raise ArgumentError, "option :to is required in button/2"
     end
 
-    form_tag(to, form) do
-      Phoenix.HTML.Form.submit(text, opts)
-    end
+    {to, form, opts}
   end
 
   defp form_options(opts, method, class) do
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -72,12 +72,8 @@ defmodule Phoenix.HTML.Link do
   end
 
   def link(text, opts) do
-    {to, opts} = Keyword.pop(opts, :to)
+    {to, opts} = pop_required_option!(opts, :to, "expected non-nil value for :to in link/2")
     {method, opts} = Keyword.pop(opts, :method, :get)
-
-    unless to do
-      raise ArgumentError, "expected non-nil value for :to in link/2, got: #{inspect to}"
-    end
 
     if method == :get do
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/link.ex
```diff
@@ -133,21 +133,44 @@ defmodule Phoenix.HTML.Link do
...
-    {form, opts} = form_options(opts, method, "button")
+  def button(opts) do
+    {to, form, opts} = extract_button_options(opts)
 
-    unless to do
-      raise ArgumentError, "option :to is required in button/2"
+    {do_option, opts} = pop_required_option(opts, :do, "option :do is required in button/1")
+
+    form_tag(to, form) do
+      Phoenix.HTML.Form.submit(opts, [do: {:safe, do_option}])
```
josevalim: This should not be needed. The EEx engine will automatically wrap it in `{:safe, ...}`. The only case it won't be wrapped is if you generating the buttons manually (say in Elixir) but then you must explicitly tag it as safe anyway (like in your tests).

test/phoenix_html/link_test.exs
```diff
@@ -67,6 +67,23 @@ defmodule Phoenix.HTML.LinkTest do
            ~s[</form>]
   end
 
+  test "button with do" do
+    csrf_token = Plug.CSRFProtection.get_csrf_token()
+
+    assert safe_to_string(button(to: "/world", class: "small", do: "<span>Hi</span>")) ==
```
josevalim: This should be `raw("<span>Hi</span>") or `{:safe, ...}` then. We also need a test this will escape unsafe output.

lib/phoenix_html/link.ex
```diff
@@ -133,21 +133,44 @@ defmodule Phoenix.HTML.Link do
...
+  defp extract_button_options(opts) do
+    {to, opts} = pop_required_option(opts, :to, "option :to is required in button/2")
+    {method, opts} = Keyword.pop(opts, :method, :post)
+
+    {form, opts} = form_options(opts, method, "button")
+
+    {to, form, opts}
+  end
+
+  defp pop_required_option(opts, key, error_message) do
```
josevalim: Good idea. I would probably add a `!` to the end of the function name. Also, in the `link` function above, in the same file, there are at least 2 occasions we could use this too. Can you please refactor that? Thank you!

drapergeek:
Thanks for the review @josevalim ! Pushed an update for all the things you mentioned.

josevalim:
I don't think we need to :show_value because it will always be or false or nil. So can simplify the pop_required_option function to what it was before. What do you think?

drapergeek:
>I don't think we need to :show_value because it will always be or false or nil. So can simplify the pop_required_option function to what it was before.

I can get behind that, I'd agree it is not especially valuable. _Technically_ that would be a change in the API so we'd need a major version bump before this is released. I don't want to hold up a release for a minor function change, we could also just not use the function in that one scenario.

josevalim:
> Technically that would be a change in the API so we'd need a major version bump before this is release

It is not a change to the API. You should never use an exception error message to decide how to act and Elixir makes that hard on purpose, especially when we have way better alternatives such as pattern matching. There is [some info here](http://elixir-lang.org/blog/2014/09/18/elixir-v1-0-0-released/). If that's your concern, don't worry. :)

drapergeek:
Sounds good to me!

drapergeek:
Updated again, thanks @josevalim !

josevalim:

PR #130, opened 2016-12-16
Add img/2 function for img tags

### LINKED ISSUE #129: image_tag?
Would you be interested in an `image_tag` function? I'm imagining a thin wrapper around content tag that takes the first argument as a binary and sets the `src`. This seems like a pretty common use case to support.

I'm happy to implement it just didn't want to do the work if you weren't interested.

drapergeek:
Fixes #129 

I wasn't sure if I should go with `img` or `img_tag`. The `button/2` function doesn't have `_tag`, nor does `form_for`. I think personally I prefer it without the unnecessary `_tag` but I could argue either way!

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -244,4 +244,20 @@ defmodule Phoenix.HTML.Tag do
     tag :meta, charset: "UTF-8", name: "csrf-token", content: get_csrf_token(),
                'csrf-param': @csrf_param, 'method-param': @method_param
   end
+
+  @doc """
+  Generates an img tag with a src.
+
+  ## Examples
+
+      img(user.photo_path)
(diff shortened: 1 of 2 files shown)


josevalim:
Looks good to me, thank you! Shall we ship it @chrismccord ?

josevalim:

chrismccord:
@josevalim just saw this. I'm not sure to be honest since we have turned down this feature several times already. Your sanitization concerns are already handled by `tag(:img, src: ...)` no? Or are you worried that folks will not realize they can call it his way and instead hardcode `"> `? I'm not opposed to adding it, but trying to understand the benefits from the issue discussion because as written we aren't doing anything with static_path and `tag` gives us sanitization so just looking like a convenience function to me. May or may not be okay

Issue #140, opened 2017-01-16
Got error "no function clause matching in Phoenix.HTML.Form.input_value/2"


dgamidov:
Hello!

After updating phoenix_html from 2.8.0 to 2.9.2 I got "no function clause matching in Phoenix.HTML.Form.input_value/2" for my checkbox (I used string as a field):

```
iex(21)> Phoenix.HTML.Form.checkbox(:checkbox, "1", value: true)
** (FunctionClauseError) no function clause matching in Phoenix.HTML.Form.input_value/2
    (phoenix_html) lib/phoenix_html/form.ex:288: Phoenix.HTML.Form.input_value(:checkbox, "1")
    (phoenix_html) lib/phoenix_html/form.ex:699: Phoenix.HTML.Form.checkbox/3
```

Integer does not work, too:

```
iex(21)> Phoenix.HTML.Form.checkbox(:checkbox, 1, value: true)
** (FunctionClauseError) no function clause matching in Phoenix.HTML.Form.input_value/2
    (phoenix_html) lib/phoenix_html/form.ex:288: Phoenix.HTML.Form.input_value(:checkbox, 1)
    (phoenix_html) lib/phoenix_html/form.ex:699: Phoenix.HTML.Form.checkbox/3
```

Atom works fine:

```
iex(21)> Phoenix.HTML.Form.checkbox(:checkbox, :test, value: true)
{:safe,
 [[60, "input",
   [[32, "name", 61, 34, "checkbox[test]", 34],
    [32, "type", 61, 34, "hidden", 34], [32, "value", 61, 34, "false", 34]],
   62],
  [60, "input",
   [[32, "checked", 61, 34, "checked", 34],
    [32, "id", 61, 34, "checkbox_test", 34],
    [32, "name", 61, 34, "checkbox[test]", 34],
    [32, "type", 61, 34, "checkbox", 34], [32, "value", 61, 34, "true", 34]],
   62]]}
```

I have checked function Phoenix.HTML.Form.input_value/2:

```
iex(19)> Phoenix.HTML.Form.input_value(:test, :"1")
nil

iex(20)> Phoenix.HTML.Form.input_value(:test, 1)
** (FunctionClauseError) no function clause matching in Phoenix.HTML.Form.input_value/2
    (phoenix_html) lib/phoenix_html/form.ex:288: Phoenix.HTML.Form.input_value(:test, 1)

iex(20)> Phoenix.HTML.Form.input_value(:test, "1")
[...]

PR #141, opened 2017-01-16
Remove guard is_atom(field) from input_value/2 without form, add tests

dgamidov:
Avoid `(FunctionClauseError) no function clause matching in Phoenix.HTML.Form.input_value/2` if field is not an atom.

According [this discussion](https://github.com/phoenixframework/phoenix_html/issues/140).

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -299,7 +299,7 @@ defmodule Phoenix.HTML.Form do
     end
   end
 
-  def input_value(name, field) when is_atom(name) and is_atom(field),
+  def input_value(name, field) when is_atom(name),
     do: nil
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:

rafaelpc:
I'm glad I've found this - couldn't figure out why my checkboxes were raising this error. I was trying to make a `many_to_many` relationship work with checkboxes as it is in [this code](https://github.com/dgamidov/checkboxes_ex/blob/master/web/templates/product/form.html.eex#L32) but using with the `put_assoc` as described in the What's new in Ecto 2.0. Thanks, @dgamidov and @josevalim for addressing this issue.

So until a new version is released, I guess I need to point to the master branch? Is the following the right way to specify it in the `deps` function of the `mix.exs` file?

`{:phoenix_html, git: "https://github.com/phoenixframework/phoenix_html.git", branch: "master", override: true},`

Btw, I think you forgot to ignore/underscore the `field` arg because is giving this warning:
`warning: variable "field" is unused
  lib/phoenix_html/form.ex:305`

josevalim:
I just published a new version! enjoy!

rafaelpc:
Many thanks José! Fastest response ever.

PR #142, opened 2017-01-18
Add date, datetime and time input types

vortizhe:
Hi, I miss date and time input types. 

I think input[type=datetime] it's necessary for using native datepickers on modern and mobile browsers .

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -518,8 +518,8 @@ defmodule Phoenix.HTML.Form do
 
   See `text_input/3` for example and docs.
   """
-  def datetime_input(form, field, opts \\ []) do
-    generic_input(:datetime, form, field, opts)
+  def datetime_local_input(form, field, opts \\ []) do
+    generic_input(:'datetime-local', form, field, opts)
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:
I believe `type="datetime"` is no longer supported and removed from most browsers in favor of local_datetime.

nekath:
It looks like datetime is no longer supported.
see: https://github.com/whatwg/html/issues/336

`datetime-local` is the correct type
see: https://www.w3.org/TR/2012/WD-html-markup-20120315/input.datetime-local.html

vortizhe:
True, isn't deprecated yet, at least on chrome but `type=datetime-local` is more future proof.

I update it.

josevalim:
Thank you! I would also like to see tests where the values are the Time, Date and NaiveDateTime structs. And I would probably rename datetime_local_input to naive_datetime_input.

josevalim:
Ping? We should standardize around Elixir's calendar types at this point.

florish:
@vortizhe Thanks for opening this, I was looking for a `date_input` function in the docs, but it turns out, there isn't one yet. Is it still on your list to finish this PR? If not, I could pick it up and proess @josevalim 's feedback on the `naive_datetime_input` rename and additional tests.

PR #147, opened 2017-03-03
Add support for generating a multiple options select to `select/4`

christos:
As a Rails developer, I was surprised when I tried to generate a multiple options select with `select f, :categories, @categories, multiple: true` and it didn't work.

This PR adds support for doing just that. It also replaces the implementation of `multiple_select/4` to call `select/4` with the applicable arguments instead.

My personal preference would be to deprecate `multiple_select` in favour of `select/4` thus reducing the cognitive dissonance especially for Rails developers many of whom are moving to Phoenix. To that end I've added a (commented out) deprecation block, as I've seen done in other parts of the project. If the maintainers decide that they don't want to deprecate, let me know and I can remove that commit.

I am an Elixir newbie, so even though I've spent a fair amount of time trying to avoid `if/else` I couldn't see the functional way of implementing this without re-writting a large portion of the helpers. I am open to feedback though. :)

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -775,6 +775,12 @@ defmodule Phoenix.HTML.Form do
           <option value="user">User</option>
           </select>
 
+      select(form, :roles, ["Admin": 1, "Power User": 2], multiple: true, selected: [1])
+      #=> <select id="user_roles" name="user[roles][]">
+          <option value="1" selected="selected" >Admin</option>
+          <option value="2">Power User</option>
+          </select>
+
   If you want to select an option that comes from the database,
   such as a manager for a given project, you may write:
 
@@ -799,6 +805,11 @@ defmodule Phoenix.HTML.Form do
 
   ## Options
 
+    * `multiple` - When set to any value other than `false` or `nil` the
+      generated select is a multiple select. Unless explicitly provided,
+      the name attribute has `[]` appended to it, in order to signal to
+      the params parser that the selected values should go in an array.
+
     * `:prompt` - an option to include at the top of the options with
       the given prompt text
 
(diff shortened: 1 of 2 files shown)


josevalim:
> My personal preference would be to deprecate multiple_select in favour of select/4 thus reducing the cognitive dissonance especially for Rails developers many of whom are moving to Phoenix. 

Sorry but you haven't given any good argument for deprecating `multiple_select` besides "it is confusing for Rails developers moving to Phoenix". That's not a valid reason. We design APIs as we think they work better, regardless of what Rails does, and having multiple_select is better because the value of the inputs and options change when you are using `multiple_select` as well considerable aspects of the output (such as having the name changed to `foo[]`).

If you believe multiple_select is not a good API, can you please ellaborate?

christos:
Jose, you are right. "Confusing to Rails developers" is not an argument. My bad. 

Thanks for taking the time to properly argue your decision. Despite the fact that your argument is very convincing, I'll have a go at explaining my rationale better.

Lets look at the `datetime_select/date_select` functions that generate a group of related `selects`; not a single HTML element.

Their API makes sense to me because it does a lot more than just generate a single HTML element. It generates a group of elements, and provides options for the programatic generation of date range `options` in those elements.

I would argue that the names should be more descriptive, like `datetime_selects` or `datetime_picker`, but that's besides the point. 

On the other hand, `select` and `multiple_select` only generate a single `select` with the most visible difference being that `multiple` boolean HTML attribute. And the `[]` trick...

The problem I see is that they can't be used interchangeably. 

Wheres `email_input()` and `text_input(type: 'email')` for example would generate the same HTML element, `select(multiple: 'multiple')` and `multiple_select()` won't. 

Do you see what I mean? The `*_input` naming convention "decorates" the `input` element with HTML attributes but the `*_select` convention changes the generated element(s) with more than just `attributes`.

And why do I care apart from OCD, you ask? 

Basically because while developing an app, I wrote `select f, :categories, @categories, multiple: ""` and a multiple select was generated, with no warning. It even worked correctly as long as I only selected one option.

However, when selecting two or more options, the rest of the options would "quietly" disappear even though I could see them being sent by the browser in the encoded form data. Lots of hair-pulling later, I discovered `multiple_select`. 😄

Maybe the real issue is that `select` allows you to pass a `multiple` option but the params parsing code won't correctly parse form-data fragments like `post[categories]=elixir&post[categories]=phoenix`. It only keeps the first occurence of the key, instead of adding them to an array.

In my opinion, the above behaviour violates the "principle" of least surprise. 

Maybe my solution was not the correct one. Maybe `select` should just spit a warning about using `multiple_select` instead and remove the `multiple` attribute, if passed as an option. I'd be happy to PR that if I've managed to convince you in any way. 😇

josevalim:
> In my opinion, the above behaviour violates the "principle" of least surprise.

The principle of least surprise changes is based on your personal previous experiences and you were only surprised because you had the wrong expectation it would behave exactly like Rails. Which is not true for Elixir nor Phoenix.

And I believe you reading too much in the naming structure. For example, check `radio_button` which end-up using inputs and it is not named `radio_input`. We could make a case that `multiple_select` could be better named but making it an option for `select` is absolutely not the way to solve it.

If the goal is to avoid confusion, we should probably mention in the `select` documentation that there is a `multiple_select` function. If that's the case, could you please send a documentation PR then? Thank you!

christos:
@josevalim Understood. 

Documentation PR created.

I am also leaving this here, for future reference:

screen shot 2017-03-03 at 14 34 52

PR #154, opened 2017-04-17
Use let/const statements to define JS variables

roperzh:
Hello :)

Since the code is transpiled to an ES5 version, we can safely
use some ES6 goodness in the source file.

Please note that the ES5 file remains equal even after running Brunch.


chrismccord:
Please see other discussions around let/const. Thanks!

Issue #151, opened 2017-03-25
`method: :delete` link and inline form that it generates


GBH:
Phoenix v.1.3-rc

Right now link with `method: delete` will generate a form that is getting inlined right next to it. This creates a problem if you put that delete link somewhere inside an existing form. This will effectively nest forms. Invalid html aside, it will trigger delete action when you just want to submit your original form. Not expected. 

I believe Rails used to have the same mechanism before v3. Now it creates form on click and puts in somewhere at the end of the `<body>` so it doesn't interfere with things. See: https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L210 (Rails5 moved from jquery, but it's the same idea)

Thoughts?

josevalim:
@GBH :+1: for this change since we depend on JavaScript anyway. If we do this though, we should probably stop using data-submit="parent" and also rely on data-method="...".

PR please?

almielczarek:
If no one is working on this, I think I know what to do to to correct this behavior. Let me know!

GBH:
@almielczarek it's all yours. I'm not much of a javascript person.

josevalim:
Fixing in favor of the PR.

ckruse:
This is a pretty breaking change. I relied on the inline forms. I guess there is no way to get back the old behavior but write my own helper with the same API as the old one?

josevalim:
@ckruse you can also depend on the previous version explicitly for now. Is there any reason why you would prefer the form version? Given it has limitations regarding nesting?

ckruse:
Pinning the version seems like a problem in the long run.

I relied on the inline form solution because it easily allowed me to create buttons for non-get requests without relying on JS. I develop everything with progressive enhancement in mind, this also means: if you don't have JS the application should still at least continue to work.

The assumption that the user don't has JS comes in different forms: he may have turned it off (NoScript extension?), he may have a bad internet connection and thus it hasn't been loaded yet, the client may be incapable of features like `querySelector()` et al, etc, pp.

It also is not true that one already has to have JS to use applications built with Phoenix, I am maintaining an application where JS is optional. It behaves different and is not as nice to use as with JS, but it works.

josevalim:
Thanks for the explanation and apologies for not considering those use cases. One last question, didn't the previous version use data-submit=parent though which required javascript?

ckruse:
[...]

PR #157, opened 2017-04-26
Fixed behavior of links (#151)

almielczarek:
Links with a `:method` other than `:get` no longer create inline forms.

This caused issues with nesting as layed out in #151. Now the forms are
generated, appended to the end of `<body>`, and submitted via Javascript
when the link is clicked.

The new links -- with the options set -- are rendered as follows: `<a
href="#" data-csrf="#{csrf_token}" data-method="#{method}"
data-to="#{to}">#{text}</a>`.

As part of this change `data-submit="parent"` is deprecated, and
`data-method="#{method}"` and `data-csrf="#{csrf_token}"` are introduced.

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -1,41 +1,38 @@
-'use strict';
+"use strict";
 
-function isLinkToSubmitParent(element) {
-  var isLinkTag = element.tagName === 'A';
-  var shouldSubmitParent = element.getAttribute('data-submit') === 'parent';
+(function() {
+  function buildHiddenInput(name, value) {
+    var input = document.createElement("input");
+    input.type = "hidden";
+    input.name = name;
+    input.value = value;
+    return input;
+  }
 
-  return isLinkTag && shouldSubmitParent;
-}
-
-function getClosestForm(element) {
-  while (element && element !== document && element.nodeType === Node.ELEMENT_NODE) {
-    if (element.tagName === 'FORM') {
-      return element;
+  function handleLinkClick(link) {
+    var message = link.getAttribute("data-confirm");
+    if(message && !window.confirm(message)) {
+        return;
     }
-    element = element.parentNode;
-  }
-  return null;
-}
 
-function didHandleSubmitLinkClick(element) {
-  while (element && element.getAttribute) {
-    if (isLinkToSubmitParent(element)) {
-      var message = element.getAttribute('data-confirm');
-      if (message === null || confirm(message)) {
-        getClosestForm(element).submit();
-      }
-      return true;
-    } else {
-      element = element.parentNode;
+    var to = link.getAttribute("data-to"),
+        method = buildHiddenInput("_method", link.getAttribute("data-method")),
+        csrf = buildHiddenInput("_csrf_token", link.getAttribute("data-csrf")),
+        form = document.createElement("form");
+
+    form.method = "post";
+    form.action = to;
+    form.style.display = "hidden";
+
+    form.appendChild(csrf);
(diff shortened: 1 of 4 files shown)


web/static/js/phoenix_html.js
```diff
@@ -1,38 +1,40 @@
...
-  return null;
-}
 
-function didHandleSubmitLinkClick(element) {
-  while (element && element.getAttribute) {
-    if (isLinkToSubmitParent(element)) {
-      var message = element.getAttribute('data-confirm');
-      if (message === null || confirm(message)) {
-        getClosestForm(element).submit();
+  window.onload = function() {
```
josevalim: Using `window.onload` means that new elements added dynamically via javascript won't get the new `handleLinkClick`. I think we need to keep the `window.addEventListener` approach, no?

almielczarek: You're correct! I'll dig back into this after work today. If you see anything else let me know.

almielczarek: I'm also going to see if I can clean up the diffs for the Javascript. Didn't see how unreadable they were.

almielczarek: I updated this to use `window.addEventListener` again. Sorry, I could not figure out how to make the diffs read nicer for the Javascript files.

almielczarek: Also, should the behavior of the button type links be changed as well, as those also currently create inline forms?

josevalim:
Thank you!

Btw, have you tested this on which browsers? I see you are using functions like `dataset` but they are relatively recent. For example, they require IE 11. @chrismccord, what is the minimal IE version we aim to support in Phoenix?

And if buttons are also using forms, we should probably fix them too.

almielczarek:
I did check it on IE, but it was version 11, I can fix that. I think we have some old XP machines mothballed at work so I can test everything else on my lunch break. I figured we should get the buttons too, so i already started that. If you see anything else, let me know! I may be out of town tonight, but I could definitely get it done tomorrow.

almielczarek:
I'm pretty sure `addEventListener` is only supported in IE 9+ too, so if we are looking to support older versions, we need to check if the function is available and if not, use `attachEvent`.

priv/static/phoenix_html.js
```diff
@@ -1,41 +1,37 @@
...
-    if (isLinkToSubmitParent(element)) {
-      var message = element.getAttribute('data-confirm');
-      if (message === null || confirm(message)) {
-        getClosestForm(element).submit();
-      }
-      return true;
-    } else {
-      element = element.parentNode;
+  function handleLinkClick(link) {
+    if(link.dataset["confirm"] == "true" && !window.confirm("Are you sure?")) {
```
kelvinst: I guess for this we could stay with the previous behavior, don't we?

It's better if we keep the message on the HTML because then we can use `<%= gettext("Are you sure?") %>` for translated messages.

josevalim: Yes. The text should not be in the JS. Good catch!

almielczarek: Noted. I'll revert to the old implementation.

chrismccord:
We should support IE9+

almielczarek:
At this point in time, I do not have access to any computers running anything below IE11. If anyone would be willing to test in those browsers it would be appreciated!

almielczarek:
If you are testing from an existing project you will need to clean your deps and node_modules, otherwise the old Javascript will be loaded. I found that to be an issue for me earlier.

josevalim:

Issue #123, opened 2016-11-18
Merge keys on content_tag


Grafikart:
The function `content_tag` accepts a list as a second argument. We could expect keys to be merged to avoid bad markup. Here is a doctest example of what could be expected.

```
      iex> opts = [class: 'demo']
      ...> content_tag :p, class: opts ++ [class: 'demo2'], class: "test2" do
      ...>   "Hello"
      ...> end
      {:safe, [60, "p", " class=\"demo demo2\"", 62, "Hello", 60, 47, "p", 62]}
```

josevalim:
Closing in favor of the PR.

PR #158, opened 2017-05-17
#123 merge keys on content_tag

Kr00lIX:
Combining html attributes with the same keys

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -80,7 +80,11 @@ defmodule Phoenix.HTML.Tag do
 
   defp tag_attrs([]), do: []
   defp tag_attrs(attrs) do
-    for {k, v} <- attrs do
+    merged_attrs = Enum.reduce(attrs, %{}, fn({k, v}, acc) ->
+      Map.update(acc, k, v, fn(current_value) -> [current_value, ' ', v] end)
+    end)
+
+    for {k, v} <- merged_attrs do
       [?\s, k, ?=, ?", attr_escape(v), ?"]
     end
   end
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -7,6 +7,7 @@ defmodule Phoenix.HTML.Tag do
   import Plug.CSRFProtection, only: [get_csrf_token: 0]
 
   @tag_prefixes [:aria, :data]
+  @tag_concat [:class]
   @csrf_param "_csrf_token"
   @method_param "_method"
 
@@ -80,11 +81,7 @@ defmodule Phoenix.HTML.Tag do
 
   defp tag_attrs([]), do: []
   defp tag_attrs(attrs) do
-    merged_attrs = Enum.reduce(attrs, %{}, fn({k, v}, acc) ->
-      Map.update(acc, k, v, fn(current_value) -> [current_value, ' ', v] end)
-    end)
-
-    for {k, v} <- merged_attrs do
+    for {k, v} <- attrs do
       [?\s, k, ?=, ?", attr_escape(v), ?"]
     end
   end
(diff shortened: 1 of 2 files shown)


michalmuskala:
If we really want to do this, I think it should be done in the `build_attrs` as a single pass over the attributes - we could build a map instead of the list there.

josevalim:
Thank you for the PR!

The issue with this approach is that it is not a generally applicable rule. We can see how data and class are handled in very different ways and each attribute will add its own logic (or require overriding, such as `:id`).

Kr00lIX:
If we still want to do it for the `class`, I can rewrite this code.

I will add this changes to `build_attrs` method and change accumulator to a map

Kr00lIX:
@michalmuskala is it ok now?

lib/phoenix_html/tag.ex
```diff
@@ -99,22 +100,31 @@ defmodule Phoenix.HTML.Tag do
...
-
-  defp build_attrs(_tag, [], acc),
-    do: acc |> Enum.sort |> tag_attrs
+  defp build_attrs(tag, attrs), do: build_attrs(tag, attrs, %{})
+
+  defp build_attrs(_tag, [], acc) do
+    Enum.map(acc, fn
+      {k, v} when is_atom(k) -> {Atom.to_string(k), v}
+      {k, v} when is_bitstring(k) -> {k, v}
+    end)
```
michalmuskala: Why is this conversion needed, it wasn't there before, I believe?

lib/phoenix_html/tag.ex
```diff
@@ -123,12 +133,13 @@ defmodule Phoenix.HTML.Tag do
...
   end
   defp build_attrs(tag, [{k, v}|t], acc) do
-    build_attrs(tag, t, [{dasherize(k), v}|acc])
+    build_attrs(tag, t, Map.put(acc, dasherize(k), v))
   end
 
   defp dasherize(value) when is_atom(value),   do: dasherize(Atom.to_string(value))
   defp dasherize(value) when is_binary(value), do: String.replace(value, "_", "-")
 
+
```
michalmuskala: ✂️

lib/phoenix_html/tag.ex
michalmuskala: If anything, we should test for `is_binary`, arbitrary bitstrings would probably crash somewhere down the road

lib/phoenix_html/tag.ex
```diff
@@ -99,22 +100,31 @@ defmodule Phoenix.HTML.Tag do
...
+  defp build_attrs(_tag, [], acc) do
+    Enum.map(acc, fn
+      {k, v} when is_atom(k) -> {Atom.to_string(k), v}
+      {k, v} when is_bitstring(k) -> {k, v}
+    end)
+    |> Enum.sort |> tag_attrs
+  end
+  defp build_attrs(tag, [{k, v}|t], acc) when k in @tag_concat do
+    acc = Map.update(acc, k, v, fn(cv) -> [cv, ' ', v] end)
```
michalmuskala: We can use `[cv, ?\s | v]` as a more memory-efficient way to store this as iodata.

lib/phoenix_html/tag.ex
Kr00lIX: Oops, there was my mistake. Thank you

lib/phoenix_html/tag.ex
Kr00lIX: Thanks, I will know that for the future

Kr00lIX:
Thanks for the review, I fixed it

michalmuskala:
Now it's up to somebody that could actually merge the PR 😆

Kr00lIX:
Thanks for the review @josevalim ! Pushed an update for all the things you mentioned.

josevalim:
This seems to be a backwards incompatible solution if some attributes can be repeated. Is there any attribute that can behave as such?

kelvinst:
Well, never heard about the possibility to set the same attribute multiple times in HTML. But also, I could not find anything on HTML specification that defines what happens then. :/

Testing a little bit here, on the cases like:

```html
<a href="/portfolios" href="/portfolios2" id="this-el">Portfolios</a>
```

When you get the value of href with:

```js
document.getElementById("this-el").href
// => "http://localhost:4000/portfolios"
```

Looks like the browser uses the first informed value of the property, which is actually the inverse of what you're doing when you merge the new attribute on the current map.

I particularly like this PR idea because it makes sense to our heads, but considering it breaks the default browser behavior, I think this ideas is not as good as it looks like.

josevalim:
Sorry for the delay on this. Since I am still not convinced we want to generally perform this change, I will go ahead and close this. Thank you!

PR #159, opened 2017-05-19
Change sigil_e delimiter for improved documentation rendering

eellson:
As the docs use `~S"""..."""` by convention, and the example sigil shown
in these docs was also trying to use `"""` as its delimiter, a `\` had
been inserted to avoid breaking out of the original string.

This of course meant the `\` was rendering in the documentation, which
seems less than ideal (you can see this in action at
https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#datetime_select/3).

Fixed by using an alternative delimiter. I picked `'''`, which is
hopefully not too controversial?

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -951,7 +951,7 @@ defmodule Phoenix.HTML.Form do
 
   ## Datetime
 
-  @doc ~S"""
+  @doc ~S'''
   Generates select tags for datetime.
 
   ## Examples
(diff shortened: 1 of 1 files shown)


michalmuskala:
I would rather change the `~S` sigil delimiter and leave "pure" `"""` in the docs

josevalim:
Yes, can we please swap the `"""` and `'''`? :)

eellson:
done :)

josevalim:

PR #161, opened 2017-06-24
Fix ArgumentError in Phoenix.HTML.Form.textarea/3

little-bobby-tables:
Passing a type other than binary to `Phoenix.HTML.Form.textarea/3` as the `value` option raises an error:

```
** (ArgumentError) lists in Phoenix.HTML and templates may only contain integers representing bytes, binaries or other lists, got invalid entry: ...
```

**Expected behavior:** the function should accept any data structure that implements the `Phoenix.HTML.Safe` protocol.

This seems like an easy fix to me, but I might be missing something important, so any feedback is appreciated :smiley:


josevalim:

PR #162, opened 2017-06-24
Optimize buffer building in Phoenix.HTML.Engine

michalmuskala:
It was quite frequent when compiling the template to generate nested iodata
composed entirely of static data. For example:

    [[["", "\n  "], "456"], "\n"]

In this PR, this kind of things are simplified at compile-time to a single
binary. This improves both the speed and memory usage of rendering templates.
This also significantly simplifies the AST and allows later compiler passes
to optimize even more - for example entirely eliminating the safe tuples in
some cases.

### DIFF
--- a/lib/phoenix_html.ex
+++ b/lib/phoenix_html.ex
@@ -63,33 +63,33 @@ defmodule Phoenix.HTML do
   @typedoc "May be safe or unsafe (i.e. it needs to be converted)"
   @type unsafe  :: Phoenix.HTML.Safe.t
 
-  @doc """
+  @doc ~s'''
   Provides `~e` sigil with HTML safe EEx syntax inside source files.
 
   Raises on attempts to interpolate with `\#{}`, so `~E` should be preferred.
 
-      iex> ~e"\""
+      iex> ~e"""
       ...> Hello <%= "world" %>
-      ...> "\""
-      {:safe, [[["" | "Hello "] | "world"] | "\\n"]}
+      ...> """
+      {:safe, "Hello world\\n"}
 
-  """
+  '''
   defmacro sigil_e(expr, opts) do
     handle_sigil(expr, opts, __CALLER__.line)
   end
 
-  @doc """
+  @doc ~s'''
   Provides `~E` sigil with HTML safe EEx syntax inside source files.
 
   Does not raise on attempts to interpolate with `\#{}`, but rather shows those
   characters literally, so it should be preferred over `~e`.
 
-      iex> ~E"\""
+      iex> ~E"""
       ...> Hello <%= "world" %>
-      ...> "\""
-      {:safe, [[["" | "Hello "] | "world"] | "\\n"]}
+      ...> """
+      {:safe, "Hello world\\n"}
 
-  """
+  '''
   defmacro sigil_E(expr, opts) do
     handle_sigil(expr, opts, __CALLER__.line)
   end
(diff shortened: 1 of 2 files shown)


lib/phoenix_html.ex
```diff
@@ -63,33 +63,33 @@ defmodule Phoenix.HTML do
...
 
   Raises on attempts to interpolate with `\#{}`, so `~E` should be preferred.
 
-      iex> ~e"\""
+      iex> ~e"""
       ...> Hello <%= "world" %>
-      ...> "\""
-      {:safe, [[["" | "Hello "] | "world"] | "\\n"]}
+      ...> """
+      {:safe, "Hello world\\n"}
```
michalmuskala: I'm not sure about this example now. It might be a good idea to show some more complex scenario that won't produce a single binary so that people don't assume it will always return a string.

michalmuskala:
I was trying this with hexpm and it looks like there are some issues, I need to see what's going on.

lib/phoenix_html/engine.ex
```diff
@@ -16,7 +16,39 @@ defmodule Phoenix.HTML.Engine do
...
+  defp compact_safe(other) do
+    other
+  end
+
+  defp compact_static([binary1, binary2 | rest])
+       when is_binary(binary1) and is_binary(binary2) do
+    compact_static([binary1 <> binary2 | rest])
+  end
+  defp compact_static([{_, _, _} = ast | rest]) do
+    compacted = Macro.prewalk(ast, &compact_safe/1)
```
josevalim: I am a bit worried about this tree traversal and lack of tests. Why do we need to traverse the tree? Shouldn't we instead try to compact as we go in handle_text instead of traversing a tree later on. Imagine I have this code:

    <%= Foo.bar ["foo", "bar"] %>

I have the impression your code would wrongly compact it.

michalmuskala: No, we only compact inside `{:safe, list}` tuples. But if you had:

```
<%= Foo.bar {:safe, ["foo", "bar"]} %>
```

It would indeed be compacted. The AST traversal was to solve a case of nested "literals", for example:

```
<%= if @foo do %>
  <%= 456 %>
<% end %>
```

would generate something like:

```
case if(fetch!(assigns, :foo), do: {:safe, [["" | "456"] | "\n"]}, else: []) do
  {:safe, safe} -> safe
  binary when is_binary(binary) -> Plug.HTML.escape(binary)
  other -> Phoenix.HTML.Safe.to_iodata(other)
end
```
So we need a way to "get inside" the case. But you're right - it might be enough to do the modifications at the buffer building stage instead of doing them after.

michalmuskala: As to the tests, I haven't explored how good is the coverage.

josevalim: Can you please explore compacting as we build the AST? That would ease my mind. Unfortunately anyone can create a `{:safe, ...}` tuple.

josevalim: Ping.

michalmuskala:
I have to admit I'm not 100% sure how to do it in a safe way.

josevalim:
@michalmuskala I think you just need to be more conservative in your implementation. Only build the buffer for the AST that you see during the engine. It should be buffered at the "root level" - never prewalk.

michalmuskala:
I spent quite some time and with every approach I tried, I either faced some issue or didn't get the expected benefits. I'm going to close this.

PR #163, opened 2017-06-29
Add support for Bootstrap modals

marceloomens:
I'm no fan of the blocking browser confirmation dialogues currently used for delete confirmation by phoenix_html. I've therefore implemented the delegation of this dialogue to a [Bootstrap modal](https://getbootstrap.com/javascript/#modals) if (and only if) jQuery is available and a suitable DOM element is present.

This seems reasonable to me since Phoenix ships with Bootstrap out of the box.

Example Bootstrap modal:

bs-modal

Delegation of the confirmation dialogue to a Bootstrap modal requires only that one adds some HTML to the template that (roughly) conforms to the following code:

```html
<div class="modal fade" id="phoenix-bs-modal" tabindex="-1" role="dialog">
  <div class="modal-dialog" role="document">
    <div class="modal-content">
      <div class="modal-header">
        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
        <h4 class="modal-title">Modal title</h4>
      </div>
      <div class="modal-body">
        <p>One fine body&hellip;</p>
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Save changes</button>
      </div>
    </div><!-- /.modal-content -->
  </div><!-- /.modal-dialog -->
</div><!-- /.modal -->
```

Important are a `div.modal#phoenix-bs-modal` element and a `.btn-primary` child element. The message passed in via the data-confirm attribute on the delete link is forwarded to the modal body.

I'm proposing this pull request as I believe that this is as some functionality that others could be interested in. Please note that I'm no Javascript expert. If you decide to pull in these changes, than a code review by someone more capable than I am would be a very good idea ;-)

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -1,5 +1,9 @@
 'use strict';
 
+/**
+* The underneath has been adapted from:
+* https://github.com/phoenixframework/phoenix_html/blob/51b0866afb3907cda652b94e8be77bc6929608d7/priv/static/phoenix_html.js
+*/
 function isLinkToSubmitParent(element) {
   var isLinkTag = element.tagName === 'A';
   var shouldSubmitParent = element.getAttribute('data-submit') === 'parent';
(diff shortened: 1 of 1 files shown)


josevalim:
Phoenix ships with bootstrap rather as a mean to an end: which is provide a nice default looking, but we don't integrate with bootstrap in any way. Thank you!

PR #166, opened 2017-07-25
Allow specifying a custom CSRF token generator

britto:
**Work In Progress**

The current implementation is tied up to Plug's CSRF protection code. We need some flexibility around the form-related helpers to, for example, generate a CSRF token input compatible with other frameworks or generate a blank token, so the final HTML output is cacheable.

#### TODO

* [x] Docs
* [x] Tests
* [x] Refactor

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -57,6 +57,14 @@ defmodule Phoenix.HTML.Link do
   or use the above data attributes, `Phoenix.HTML` relies
   on JavaScript. You can load `priv/static/phoenix_html.js`
   into your build tool.
+
+  ## CSRF Protection
+
+  By default, CSRF tokens are generated through `Plug.CSRFProtection`. You
+  can customize the CSRF token generation by configuring your own MFA:
+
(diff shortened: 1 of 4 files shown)


lib/phoenix_html/link.ex
```diff
@@ -135,13 +135,18 @@ defmodule Phoenix.HTML.Link do
...
-    if csrf_token = csrf_token? && Plug.CSRFProtection.get_csrf_token() do
+    if csrf_token = csrf_token? && get_csrf_token() do
       {[csrf: csrf_token], opts}
     else
       {[], opts}
     end
   end
 
+  defp get_csrf_token do
+    generator = Application.get_env(:phoenix_html, :csrf_token_generator, Plug.CSRFProtection)
```
josevalim: I think it would be better if it was a MFA (module-function-args):

    {mod, fun, args} = Application.fetch_env(:phoenix_html, :csrf_token_generator)
    apply(mod, fun, args)

Also note that you don't need a default here. Instead, in the "mix.exs" file under `def application`, add this:

    env: {Plug.CSRFProtection, :get_csrf_token, []}

and now you can expect it to always be available.

josevalim:

PR #169, opened 2017-07-30
Wrap button/link with data-method="get" in form with same method

thmzlt:
With this change, creating a button/link with `method: "get"` causes the generated form to issue a corresponding GET request.


josevalim:

PR #170, opened 2017-08-14
Correct datetime_builder/5 for [9th day of month] option tag

rickclare:
PR #149 introduced a regression in `datetime_builder/5` - the HTML in the `<option>` tag for the "9th day of the month" is rendered inconsistently when compared to the option tags for days 1 to 8.

Current behaviour:

```html
...
<option value="7">07</option>
<option value="8">08</option>
<option value="9">9</option>
<option value="9">10</option>
...
```

Corrected behaviour (i.e. "9" should be "09")

```html
...
<option value="7">07</option>
<option value="8">08</option>
<option value="9">09</option>
<option value="9">10</option>
...
```

I have corrected the source-code in `lib/phoenix_html/form.ex`, and also added to some of the existing test scenarios in `test/phoenix_html/form_test.exs` to test the solution.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1148,7 +1148,7 @@ defmodule Phoenix.HTML.Form do
   ]
 
   map = &Enum.map(&1, fn i ->
-    pre = if i < 9, do: "0"
+    pre = if i < 10, do: "0"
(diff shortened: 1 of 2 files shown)


josevalim:

PR #172, opened 2017-09-01
Whitelist accepted URI schemes

GriffinMB:
This is a PR to allow conversation to continue, but isn't yet complete. The current implementation only raises for "link", but I suspect we will also want to do so for "button"?  

I went with a [whitelist approach](https://security.stackexchange.com/questions/148428/which-url-schemes-are-dangerous-xss-exploitable). And I also added a start to the tuples allowing javascript or data URIs. It is currently only:

```
defp invalid_destination?(to) when is_tuple(to), do: false
```

Do we want to implement the functionality to transform `{:javascript, val}` within the `content_tag` function? Or do we want the `link` function to transform that value?

Edit:

I do think transforming w/in link could be nice, because that would allow us to clear the invalid URL instead of raising (which becomes a DoS as opposed to an XSS).

Something to the effect of...

```
...
to = dest(to)
...
for scheme <- @valid_uri_schemes do
  defp dest(unquote(scheme) <> _ = to), do: to
end
defp dest({scheme, to}) when is_atom(scheme) do
  if String.starts_with?(to, "#{scheme}:"), do: to
end
defp dest(to) when is_binary(to) do
  if !String.contains?(to, ":"), do: to
end
... 
```

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -66,6 +66,11 @@ defmodule Phoenix.HTML.Link do
       config :phoenix_html, csrf_token_generator: {MyGenerator, :get_token, []}
 
   """
+  @valid_uri_schemes ["http:", "https:", "ftp:", "ftps:", "mailto:",
+                      "news:", "irc:", "gopher:", "nntp:", "feed:",
+                      "telnet:", "mms:", "rtsp:", "svn:", "tel:", "fax:",
+                      "xmpp:"]
+
   def link(text, opts)
 
   def link(opts, do: contents) when is_list(opts) do
@@ -78,6 +83,11 @@ defmodule Phoenix.HTML.Link do
 
   def link(text, opts) do
     {to, opts} = pop_required_option!(opts, :to, "expected non-nil value for :to in link/2")
+
+    if invalid_destination?(to) do
+      raise ArgumentError, "link/2 expects a valid URL or path"
+    end
+
     {method, opts} = Keyword.pop(opts, :method, :get)
 
     if method == :get do
@@ -165,4 +175,11 @@ defmodule Phoenix.HTML.Link do
 
     {value, opts}
   end
+
+  defp invalid_destination?(to) when is_tuple(to), do: false
+
+  for scheme <- @valid_uri_schemes do
+    defp invalid_destination?(unquote(scheme) <> _), do: false
+  end
+  defp invalid_destination?(to), do: String.contains?(to, ":")
 end
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -84,7 +84,7 @@ defmodule Phoenix.HTML.Link do
   def link(text, opts) do
     {to, opts} = pop_required_option!(opts, :to, "expected non-nil value for :to in link/2")
 
-    to = valid_destination!(to)
+    to = valid_destination!(to, "link/2")
 
     {method, opts} = Keyword.pop(opts, :method, :get)
 
@@ -137,7 +137,7 @@ defmodule Phoenix.HTML.Link do
     {to, opts} = pop_required_option!(opts, :to, "option :to is required in button/2")
     {method, opts} = Keyword.pop(opts, :method, :post)
 
-    to = valid_destination!(to)
+    to = valid_destination!(to, "button/2")
 
     if method == :get do
       opts = skip_csrf(opts)
@@ -176,10 +176,10 @@ defmodule Phoenix.HTML.Link do
     {value, opts}
   end
 
-  defp valid_destination!(to) do
+  defp valid_destination!(to, calling_func) do
     if invalid_destination?(to) do
       raise ArgumentError, """
-      unsupported scheme given to link/2. In case you want to link to an
+      unsupported scheme given to #{calling_func}. In case you want to link to an
       unknown or unsafe scheme, such as javascript, use a tuple: {:javascript, rest}
       """
     end
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -185,7 +185,8 @@ defmodule Phoenix.HTML.Link do
     end
 
     case to do
-      {_, dest} -> dest
+      {:safe, _} = dest -> dest
+      {scheme, dest} -> "#{scheme}:#{dest}"
       dest -> dest
     end
   end
@@ -193,11 +194,8 @@ defmodule Phoenix.HTML.Link do
   for scheme <- @valid_uri_schemes do
     defp invalid_destination?(unquote(scheme) <> _), do: false
   end
-  defp invalid_destination?({:safe, to}) do
-    invalid_destination?(to)
-  end
   defp invalid_destination?({scheme, to}) when is_atom(scheme) do
-    !String.starts_with?(to, "#{scheme}:")
+    invalid_destination?(to)
   end
   defp invalid_destination?(to) when is_binary(to) do
     String.contains?(to, ":")
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -84,9 +84,7 @@ defmodule Phoenix.HTML.Link do
   def link(text, opts) do
     {to, opts} = pop_required_option!(opts, :to, "expected non-nil value for :to in link/2")
 
-    if invalid_destination?(to) do
-      raise ArgumentError, "link/2 expects a valid URL or path"
-    end
+    to = valid_destination(to)
 
     {method, opts} = Keyword.pop(opts, :method, :get)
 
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/link.ex
```diff
@@ -78,6 +83,11 @@ defmodule Phoenix.HTML.Link do
 
   def link(text, opts) do
     {to, opts} = pop_required_option!(opts, :to, "expected non-nil value for :to in link/2")
+
+    if invalid_destination?(to) do
+      raise ArgumentError, "link/2 expects a valid URL or path"
```
josevalim: What about:

> unsupported scheme given to link/2. In case you want to link to an unknown or unsafe scheme, such as javascript, use a tuple: {:javascript, rest}

lib/phoenix_html/link.ex
```diff
@@ -165,4 +175,11 @@ defmodule Phoenix.HTML.Link do
 
     {value, opts}
   end
+
+  defp invalid_destination?(to) when is_tuple(to), do: false
```
josevalim: This check would allow this to pass: `link "Foo", to: html_escape(conn.params["bar])`. Is this the intent? If so, we need a test for it and we should probably tweak the message above by asking developers to use `html_escape(javascript)` instead of `{:javascript, rest}`.

lib/phoenix_html/link.ex
```diff
@@ -165,4 +175,11 @@ defmodule Phoenix.HTML.Link do
...
+
+  for scheme <- @valid_uri_schemes do
+    defp invalid_destination?(unquote(scheme) <> _), do: false
+  end
+  defp invalid_destination?(to), do: String.contains?(to, ":")
```
josevalim: We probably need a `is_binary(to)` guard?

lib/phoenix_html/link.ex
GriffinMB: Definitely a better error. What are your thoughts with regard to:

> I do think transforming [the value of "to"] w/in link could be nice, because that would allow us to clear the invalid URL instead of raising (which becomes a DoS as opposed to an XSS).

As-is the onus is still on the programmer to appropriately filter bad input. It's better than an active attack, but still not the friendliest. An additional benefit of transforming the "to" value is that we could properly handle `{:javascript, to}` within link.ex.

That said, I do also see the value of raising.

GriffinMB: And, if we don't want to transform the input here, where would we want to do it?

lib/phoenix_html/link.ex
GriffinMB: I believe the "to" value would already be escaped, so there wouldn't really be a need for them to `html_escape` that input. However, I do think this should *technically* work. Depending on your thoughts for transforming the "to" value vs raising, this tuple check will end up doing different things. But, if you like the transformation, I think the check would allow arbitrary schemes. So it could look something like this..

```
defp destination({scheme, to}) when is_atom(scheme) do
  if valid_destination?("#{scheme}:", to), do: to
end

defp valid_destination?(scheme, scheme <> _), do: true 
```

Consequently, `link "Foo", to: html_escape(conn.params["bar])` would work, but it would result in output like `<a href="safe:user_input">Foo</a>`.

lib/phoenix_html/link.ex
GriffinMB: Yep!

lib/phoenix_html/link.ex
josevalim: We have four different input combinations:

  1. invalid scheme and unsafe
  2. invalid scheme and safe
  3. valid scheme and unsafe
  4. valid scheme and safe

Currently we only check if a string has a valid scheme if it is unsafe. But do we want to consider `{:safe, "javascript:..."}` as a valid scheme or do we want to force the user to also take action on that?

I would say `{:safe, "javascript:..."}` is still invalid because someone may have done this: `link "Bar", to: html_escape(conn.params["foo"])` and that does not imply they support javascript. And even though html_escape is called implicitly, users can still call it if they want to.

If that's the case, then the implementation should probably do this:

```
defp invalid_destination?({:safe, to}) do
  invalid_destination?(to)
end
```

Does it make sense?

Regarding raising or not, the issue comes when the user passes `"javascript:..."` in development and then it is unclear why it does not work. So if we want to discard, we would need to at least log, but there is very likely other parts of Phoenix that would raise on invalid input, so I am not sure if it is worth going in this direction.

GriffinMB:
This PR should address all issues raised with the previous PR! Let me know if any other changes are needed.

lib/phoenix_html/link.ex
```diff
@@ -165,4 +175,32 @@ defmodule Phoenix.HTML.Link do
...
+  end
+
+  for scheme <- @valid_uri_schemes do
+    defp invalid_destination?(unquote(scheme) <> _), do: false
+  end
+  defp invalid_destination?({:safe, to}) do
+    invalid_destination?(to)
+  end
+  defp invalid_destination?({scheme, to}) when is_atom(scheme) do
+    !String.starts_with?(to, "#{scheme}:")
```
josevalim: Tiniest nitpick ever: we prefer to use `not` instead of `!` when the result is guaranteed to be a boolean. :)

lib/phoenix_html/link.ex
```diff
@@ -165,4 +175,32 @@ defmodule Phoenix.HTML.Link do
...
+  defp valid_destination!(to, calling_func) do
+    if invalid_destination?(to) do
+      raise ArgumentError, """
+      unsupported scheme given to #{calling_func}. In case you want to link to an
+      unknown or unsafe scheme, such as javascript, use a tuple: {:javascript, rest}
+      """
+    end
+
+    case to do
+      {_, dest} -> dest
```
josevalim: This will unfortunately unwrap a `safe` tuple causing the input to possibly be escaped twice (or be escaped when it should not).

lib/phoenix_html/link.ex
josevalim: Question: for an unsupported prefix, what if we automatically add the prefix instead of checking if it is the prefixed being used? So I would pass: `{:javascript, "alert(0)"}`. Wouldn't that be cleaner?

josevalim:
@GriffinMB I have added some final comments and we should be almost there. See you in Seattle!

lib/phoenix_html/link.ex
GriffinMB: Sure! Auto-prefix makes sense. My thought was that this would be an opportunity to validate that the URI conforms, but prefixing and ensuring validity of "to" should accomplish the same thing. And, you're right, it does feel cleaner.

lib/phoenix_html/link.ex
```diff
@@ -165,4 +175,30 @@ defmodule Phoenix.HTML.Link do
...
+      {scheme, dest} -> "#{scheme}:#{dest}"
+      dest -> dest
+    end
+  end
+
+  for scheme <- @valid_uri_schemes do
+    defp invalid_destination?(unquote(scheme) <> _), do: false
+  end
+  defp invalid_destination?({scheme, to}) when is_atom(scheme) do
+    invalid_destination?(to)
```
josevalim: Now I think that this code is not going to allow this: `{:javascript, "console.log({one: 1})"}` because it still checks for `:`. We also need to make sure this works: `{:javascript, {:safe, ...}}`.

Since there are so many corner cases, it may be better to do it in two steps instead of making it recursive so we also avoid `{:safe, {:javascript, ...}}` and other combinations. Maybe this:

```elixir
defp valid_destination!({:safe, to}) do
  {:safe, valid_destination_string!(to)}
end
defp valid_destination!({other, to}) do
  [other, ?:, to]
end
defp valid_destination!(to) do
  valid_destination_string!(to)
end
```

josevalim: PS: I slightly edited the code above.

GriffinMB:
Now nested schemes like `link("foo", to: {:javascript, {:safe, "alert(1)"}})` or `link("foo", to: {:safe, {:javascript, "alert(1)"}})` will work properly. And anything in the scheme tuple format will pass the invalid_destination check, just like standard schemes. 

Something like {:bitcoin, {:javascript, "alert(1)"}} will raise the following error: "lists in Phoenix.HTML and templates may only contain integers representing bytes, binaries or other lists". I thought this made more sense than adding a guard to `valid_destination!` and throwing an "unsupported scheme" error. But I'm open to the idea that that is the wrong choice!

lib/phoenix_html/link.ex
GriffinMB: > Now I think that this code is not going to allow this: {:javascript, "console.log({one: 1})"}

Oops, you're right. Maybe I should stick to security! 😅

josevalim:

josevalim:
I  just realized that the input to `link/2` is any iodata so I slightly adjusted this in a follow up commit. This definitely had way more corner cases than we anticipated :D

It also means we do not support something like `{:safe, {:javascript, "alert(1)"}}` but I can't see anybody writing this unless they build the tuples by hand, which there is absolutely no need to given the opposite works just fine.

PR #173, opened 2017-09-06
Fix expected option keyword list keys in docs

florish:
The documentation listed `:value` and `:label`, while the code expects `:key` and `:value`.


josevalim:

PR #175, opened 2017-09-22
Add date, datetime-local and time input types

florish:
Continuation of #142. Addresses some of the feedback @josevalim gave there:

* Tests added for `Date`, `NaiveDateTime`, and `Time` values passed to `*_input` functions
* Conversion of these values to valid strings following MDN [date](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date), [datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local), and [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time) specs

I'm not sure about the input value conversion code, feels a bit clumsy still, but it makes the tests pass, which was my first concern. I'm still fairly new to Elixir, feedback on code style is much appreciated here!

Feedback not (yet) addressed:

* Decided not to rename `datetime_local_input` to `naive_datetime_input`. Felt inconsistent to name this input after the _Elixir_ construct, where all other input tag methods have been named after their _HTML_ types.
* I'm not sure I understand what's needed to fulfill @josevalim 's request to "standardize around Elixir's calendar types". I possible, please provide some guidance (docs, articles, anything) and I'll try to work it out 😄

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -536,7 +536,12 @@ defmodule Phoenix.HTML.Form do
   See `text_input/3` for example and docs.
   """
   def datetime_local_input(form, field, opts \\ []) do
-    opts = Keyword.replace(opts, :value, datetime_local_input_value(Keyword.get(opts, :value)))
+    opts = case Keyword.fetch(opts, :value) do
+      {:ok, value} ->
+        Keyword.put(opts, :value, datetime_local_input_value(value))
+      :error ->
+        opts
+    end
     generic_input(:'datetime-local', form, field, opts)
   end
 
@@ -556,7 +561,12 @@ defmodule Phoenix.HTML.Form do
   See `text_input/3` for example and docs.
   """
   def time_input(form, field, opts \\ []) do
-    opts = Keyword.replace(opts, :value, time_input_value(Keyword.get(opts, :value)))
+    opts = case Keyword.fetch(opts, :value) do
+      {:ok, value} ->
+        Keyword.put(opts, :value, time_input_value(value))
+      :error ->
+        opts
+    end
     generic_input(:time, form, field, opts)
   end
 
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -540,9 +540,9 @@ defmodule Phoenix.HTML.Form do
     generic_input(:'datetime-local', form, field, opts)
   end
 
-  defp datetime_local_input_value(%{__struct__: NaiveDateTime} = value) do
-    str = to_string(value)
-    String.slice(str, 0..9) <> "T" <> String.slice(str, 11..15)
+  defp datetime_local_input_value(%NaiveDateTime{} = value) do
+    <<date::10-binary, ?\s, hour_minute::5-binary, _rest::binary>> = NaiveDateTime.to_string(value)
+    [date, ?T, hour_minute]
   end
 
   defp datetime_local_input_value(other), do: other
@@ -560,7 +560,7 @@ defmodule Phoenix.HTML.Form do
     generic_input(:time, form, field, opts)
   end
 
-  defp time_input_value(%{__struct__: Time} = value) do
+  defp time_input_value(%Time{} = value) do
     value
     |> to_string
     |> String.slice(0..4)
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -515,6 +515,42 @@ defmodule Phoenix.HTML.Form do
     generic_input(:range, form, field, opts)
   end
 
+  @doc """
+  Generates a date input.
+
+  Warning: this feature isn't available in all browsers.
+  Check `http://caniuse.com/#feat=input-datetime` for further informations.
+
+  See `text_input/3` for example and docs.
+  """
+  def date_input(form, field, opts \\ []) do
+    generic_input(:date, form, field, opts)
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -515,6 +515,59 @@ defmodule Phoenix.HTML.Form do
...
+  @doc """
+  Generates a time input.
+
+  Warning: this feature isn't available in all browsers.
+  Check `http://caniuse.com/#feat=input-datetime` for further informations.
+
+  See `text_input/3` for example and docs.
+  """
+  def time_input(form, field, opts \\ []) do
+    opts = Keyword.replace(opts, :value, time_input_value(Keyword.get(opts, :value)))
```
josevalim: Please use `Keyword.put` so we remain compatible with earlier Elixir versions.

lib/phoenix_html/form.ex
```diff
@@ -515,6 +515,59 @@ defmodule Phoenix.HTML.Form do
...
+  @doc """
+  Generates a datetime-local input.
+
+  Warning: this feature isn't available in all browsers.
+  Check `http://caniuse.com/#feat=input-datetime` for further informations.
+
+  See `text_input/3` for example and docs.
+  """
+  def datetime_local_input(form, field, opts \\ []) do
+    opts = Keyword.replace(opts, :value, datetime_local_input_value(Keyword.get(opts, :value)))
```
josevalim: Please use `Keyword.put` so we remain compatible with earlier Elixir versions.

lib/phoenix_html/form.ex
```diff
@@ -515,6 +515,59 @@ defmodule Phoenix.HTML.Form do
...
+    generic_input(:'datetime-local', form, field, opts)
+  end
+
+  defp datetime_local_input_value(%{__struct__: NaiveDateTime} = value) do
```
josevalim: maybe `%NaiveDateTime{} = value`?

lib/phoenix_html/form.ex
```diff
@@ -515,6 +515,59 @@ defmodule Phoenix.HTML.Form do
...
+    str = to_string(value)
+    String.slice(str, 0..9) <> "T" <> String.slice(str, 11..15)
```
josevalim: This function could likely be more efficiently implemented as:

```elixir
<<date::10-binary, ?\s, hour_minute::5-binary, _rest::binary>> = NaiveDateTime.to_string(value)
[date, ?T, hour_minute]
```

lib/phoenix_html/form.ex
```diff
@@ -515,6 +515,59 @@ defmodule Phoenix.HTML.Form do
...
+  Check `http://caniuse.com/#feat=input-datetime` for further informations.
+
+  See `text_input/3` for example and docs.
+  """
+  def time_input(form, field, opts \\ []) do
+    opts = Keyword.replace(opts, :value, time_input_value(Keyword.get(opts, :value)))
+    generic_input(:time, form, field, opts)
+  end
+
+  defp time_input_value(%{__struct__: Time} = value) do
```
josevalim: Please see the notes in `datetime_local_input`.

josevalim:
I have added some comments and then this looks good to go!

lib/phoenix_html/form.ex
florish: OK, will do!

lib/phoenix_html/form.ex
florish: I tried that first, but ran into this error (both for `time_input` and `datetime_local_input`:

```
$ mix test
Compiling 2 files (.ex)
.........................................................................................

  1) test datetime_local_input/3 with form (Phoenix.HTML.FormTest)
     test/phoenix_html/form_test.exs:358
     Assertion with == failed
     code:  assert safe_form(&(datetime_local_input(&1, :key))) == ~s"<input id=\"search_key\" name=\"search[key]\" type=\"datetime-local\" value=\"value\">"
     left:  "<input id=\"search_key\" name=\"search[key]\" type=\"datetime-local\">"
     right: "<input id=\"search_key\" name=\"search[key]\" type=\"datetime-local\" value=\"value\">"
     stacktrace:
       test/phoenix_html/form_test.exs:359: (test)
```

With, `Keyword.replace`, a `:value` entry is only added if it already existed before, while with `Keyword.put`, a `value: nil` is always added, overriding the value lookup from `form` in `generic_input` a few lines later:

```
Keyword.put_new(:value, input_value(form, field))
```

I couldn't find a way around this, other than adding lots of `if`s, which felt awkward. Would you have any backward-compatible suggestions?

lib/phoenix_html/form.ex
florish: See comment above!

lib/phoenix_html/form.ex
florish: @josevalim This is one approach that works, but feels a bit verbose:

```
def time_input(form, field, opts \\ []) do
    opts = case Keyword.fetch(opts, :value) do
      {:ok, value} ->
        Keyword.put(opts, :value, time_input_value(value))
      :error ->
        opts
    end
    generic_input(:time, form, field, opts)
  end
```

josevalim: Yes, you will have to do the long version for now then.

florish:
@josevalim Should be done now then!

josevalim:

Issue #171, opened 2017-08-20
Incorrect hidden_input for boolean.


jacekbogdanski:
I am not sure if it's Ecto casting or Phoenix bug. Phoenix.HTML.Form.hidden_input for boolean field produces for false: 
```html
<input type="hidden" id="struct_field" name="struct[field]" value="" />
```
and for true:
```html
<input type="hidden" id="struct_field" name="struct[field]" value="value" />
```
It gives casting validation error when creating changeset from params. Ecto can't cast string "value" to true. It's fine for "true" string.

josevalim:
That's the semantics of HTML attributes where the is not really a "true", but the key is repeated, such `disabled="disabled"`. If you want "true" / "false", then it needs to be cast to a string.

jacekbogdanski:
Thanks for fast reply. That's right but it seems inconsistent with ecto. Shouldn't it work out of the box? Now we have to manually cast booleans with ecto by custom type or provide "true" / "false" for hidden_input.

josevalim:
@jacekbogdanski we need to handle it here rather. The best solution is probably to call html_escape on the :value attribute inside generic_input.

PR #177, opened 2017-10-08
Escape value of form input tags

belaustegui:
This PR addresses issue #171, which I have experienced myself yesterday.

This pull request escapes the value option of form inputs to ensure that it gets represented properly in the generated HTML.

The HTML semantics of representing true attributes by repeating the key in the value (such as `disabled="disabled"`) cause some confusion on the value attribute, which would be represented as `value="value"` instead of `value="true"`.


josevalim:

PR #178, opened 2017-10-09
fix bug with unicode characters in escape_javascript/1 and safe_to_string/1

balexand:
I'm assuming that these functions should be able to handle unicode characters. Thanks!


josevalim:
This would be a backwards incompatible change. Phoenix engines' work on iodata, this means `233` in a list is the byte 233 and not the codepoint 233 (`é`). If you need to pass unicode data, then please use strings.

PR #181, opened 2017-10-25
improve documentation of `datetime_select` to enable users to l10n mo…

ckruse:
…nths

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1045,8 +1045,8 @@ defmodule Phoenix.HTML.Form do
   `:second`. All options given to those keys will be forwarded to the
   underlying select. See `select/4` for more information.
 
-  To specify a list of months e.g. for localization give a list of
-  `{"Name", "number"}` for `:options` to the `:month` key:
+  For example, if you are using Phoenix with Gettext and you want to localize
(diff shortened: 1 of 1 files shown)


lib/phoenix_html/form.ex
```diff
@@ -1045,6 +1045,27 @@ defmodule Phoenix.HTML.Form do
   `:second`. All options given to those keys will be forwarded to the
   underlying select. See `select/4` for more information.
 
+  To specify a list of months e.g. for localization give a list of
+  `{"Name", "number"}` for `:options` to the `:month` key:
+
+      # Assuming form contains a User schema
```
Gazler: I don't think we need this comment.

lib/phoenix_html/form.ex
```diff
@@ -1045,6 +1045,27 @@ defmodule Phoenix.HTML.Form do
   `:second`. All options given to those keys will be forwarded to the
   underlying select. See `select/4` for more information.
 
+  To specify a list of months e.g. for localization give a list of
+  `{"Name", "number"}` for `:options` to the `:month` key:
```
Gazler: I think we change the tuple to:

  {value, key}

ckruse:
Much better with your changes, @josevalim - thanks!

Issue #182, opened 2017-10-31
Cannot generate html5 boolean attribute using `tag` or `content_tag`


ijcd:
I am working with the javascript libraries that are parsing the dom and looking for html5 boolean values in the name-only format (no value) to indicate boolean. I can't seem to figure out how to create that using `tag` or `content_tag`:

The boolean attribute styles are described here:

http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes

Current output:

```
iex(12)> Phoenix.HTML.safe_to_string Phoenix.HTML.Tag.content_tag(:p, "Hello", "checked": true)
"<p checked=\"checked\">Hello</p>"
```

Desired output:
```
iex(12)> Phoenix.HTML.safe_to_string Phoenix.HTML.Tag.content_tag(:p, "Hello", "checked": true)
"<p checked>Hello</p>"
```

Would you accept a patch that would adjust the output to be as above? You could still generate the current format using a normal string value for the attribute.

josevalim:
Thanks @ijcd. I don' see a problem doing this change but I would like to know what is the browser support for boolean attributes (or HTML5). I think we should be covered but it would be good to sanity check. Can you please investigate and see if you find some information?

ijcd:
Some discussion here:

https://stackoverflow.com/questions/4139786/what-does-it-mean-in-html-5-when-an-attribute-is-a-boolean-attribute

Ideally, the document generator as a whole could somehow be aware of the document doctype (html4, html5, xhtml, etc...) and generate tags appropriately. These functions are just a text emitter though, the function caller would have to be aware of when which form was appropriate when.

About 18 tests broken when I made the change locally. (the output tests for the form helper, etc.) so people might feel the effect if they encountered the default.

I don't see a lot of compatibility information about this specific feature, but since it is basic DOM stuff I would think the lack of information (caniuse.com, html4test.com, etc) means it is well supported.

I do note that the developer.mozilla.org docs show the naked use of "checked" in their examples:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

Since it is required for input validation, this older article indicates that at least parsing for the "required" boolean attribute has been on all browsers since Safari 9.5(the laggard):

https://developer.telerik.com/topics/web-development/cross-browser-html5-form-validation-finally-now/

I assume the dom parsing side was around much longer.

ijcd:
[...]

PR #183, opened 2017-10-31
Add support for html5 boolean attributes to content_tag and tag.

ijcd:
Fixes https://github.com/phoenixframework/phoenix_html/issues/182

I also updated the helper tags so that the current output of using `checked="checked"` does not change. Direct use of these tags would change, however. I also updated tests and doctests to reflect this.

HTML5 boolean attributes can come in several forms.

These mean checked is true:
```
     <checkbox checked="checked">
     <checkbox checked="">
     <checkbox checked>
```

This means checked is false:

```
     <checkbox>
```

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -738,6 +738,11 @@ defmodule Phoenix.HTML.Form do
         opts
       end
 
+    opts =
+      Keyword.update(opts, :checked, false, fn c ->
+        boolean_attribute("checked", c)
+      end)      
+
     tag(:input, [value: value] ++ opts)
   end
 
@@ -797,6 +802,11 @@ defmodule Phoenix.HTML.Form do
       else
         opts
       end
+
+    opts =
+      Keyword.update(opts, :checked, false, fn c ->
+        boolean_attribute("checked", c)
+      end)
 
     html_escape [tag(:input, name: Keyword.get(opts, :name), type: "hidden", value: unchecked_value),
                  tag(:input, [value: checked_value] ++ opts)]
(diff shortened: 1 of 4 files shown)


lib/phoenix_html/tag.ex
```diff
@@ -89,8 +91,11 @@ defmodule Phoenix.HTML.Tag do
 
   defp tag_attrs([]), do: []
   defp tag_attrs(attrs) do
-    for {k, v} <- attrs do
-      [?\s, k, ?=, ?", attr_escape(v), ?"]
+    for a <- attrs do
+      case a do
+        {k, v} -> [?\s, k, ?=, ?", attr_escape(v), ?"]
+        {k} -> [?\s, k]
```
josevalim: We should not wrap in a tuple with one element. Just pass the value as is.

lib/phoenix_html/form.ex
```diff
@@ -738,6 +738,11 @@ defmodule Phoenix.HTML.Form do
         opts
       end
 
+    opts =
+      Keyword.update(opts, :checked, false, fn c ->
+        boolean_attribute("checked", c)
```
josevalim: Question: Why do we need this? Can't we just pass the boolean forward?

ijcd: Passing the boolean forward will mean the naked attribute version will be output instead. I added the boolean_attribute() helper to retain the current behavior for `form_for`, `select`, `radio_button`, etc. They currently emit the boolean form `checked="checked"` and I thought it would be less impactful for that not to change. That is still a valid boolean form. If we just pass the boolean forward, then many of the current tests need to be updated, and people might be surprised that the html generation changed. Either works -- this is just less disruptive, I feel.

josevalim: I see. Let's go ahead and rip the band-aid, otherwise people may be confused why some attributes work like that and others do not. :)

ijcd:
I removed the `boolean_attribute` helper and updated the tests. Would you normally squash out the changes from PR discussion or merge as is? I haven't contributed to this repo yet.

josevalim:
We can squash from the github UI now, so no worries! :D

josevalim:

PR #189, opened 2017-12-15
Optimise eex engine to build flattened iodata

michalmuskala:
I'll attach some benchmarks soon.

### DIFF
--- a/lib/phoenix_html.ex
+++ b/lib/phoenix_html.ex
@@ -68,10 +68,11 @@ defmodule Phoenix.HTML do
 
   Raises on attempts to interpolate with `\#{}`, so `~E` should be preferred.
 
-      iex> ~e"\""
-      ...> Hello <%= "world" %>
+      iex> {:safe, data} = ~e"\""
+      ...> Hello, <%= "world" %>
       ...> "\""
-      {:safe, [[["" | "Hello "] | "world"] | "\\n"]}
+      iex> IO.iodata_to_binary(data)
+      "Hello, world\\n"
 
   """
   defmacro sigil_e(expr, opts) do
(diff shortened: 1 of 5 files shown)


michalmuskala:
So I tried running some benchmarks using some templates from https://github.com/hexpm/hexpm but I couldn't actually see much of a difference - it was always within just couple percent, so I'm not sure it's worth the additional complexity.

josevalim:
@michalmuskala I agree. Although parts of this patch is probably worth keeping:

1. the improved docs
2. the use of html_escape_to_iodata

and probably a couple others.

Thoughts?

chrismccord:
As this resolves the Safari hanging issues with http2 (https://github.com/phoenixframework/phoenix/issues/2686), I think we are good to merge as is. @josevalim can you verify if we're good to go or any exclusions you want to see? Thanks!

michalmuskala:
I'm pretty sure the fact that it fixes the safari bug is entirely accidental. There must be something wrong at some other level, if sending proper iodata breaks stuff.

josevalim:
I agree. It would be nice to merge this but we need to have the Safari issue fixed regardless because we don't know other cases where it may occur.

idyll:
I think we need to add some larger tests for this. In my testing I though I was seeing some of the responses come out of order. I still haven't isolated it though...

idyll:
Cowboy 2.2.1 includes a fix for the Safari issue now!

idyll:
Okay with Cowboy 2.2.2 now out I have retested this branch and for a larger project I am seeing content come in out of order and scrambled.

If you do intend to merge this let me know and I will try to write a test that will demonstrate the issue that I am seeing.

However, with Cowboy now fixed I'm not sure this does need to be merged.

chrismccord:
@idyll so we're on the same page, can you verify that Phoenix master and cow 2.2.2 works as expected, but when you try this PR you get out of order content? If all is well on latest phoenix/phoenix_html/cowboy, then I think we can hold on this one.

idyll:
Sorry, just getting back to this now. 

Confirmed, master with Cowboy 2.2.2 is working. When I use this PR I am getting content out of order.
I haven't had time to track down why.

michalmuskala:
It's possible there's a bug in the implementation. Could you give an example template that reproduces the issue? I'm pretty sure in that case we won't even need to go through the http stack - just rendering should showcase the problem.

josevalim:
I pushed a version with similar improvements but friendlier compilation here: cf08dc8e07281bc269f0f563ae508c6efa030749

PR #191, opened 2017-12-20
Support carriage return in text_to_html

### LINKED ISSUE #190: Phoenix.HTML.Format.text_to_html/2 new line characters
As per docs we can use [`text_to_html/2`](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Format.html#text_to_html/2) to transform `\n\n` to paragraphs:

```
iex> text_to_html("Hello\n\nWorld") |> safe_to_string
"<p>Hello</p>\n<p>World</p>\n"
```

https://github.com/phoenixframework/phoenix_html/blob/0bb681d4e98b61a76dc03ff7a438a1fa64facd5d/lib/phoenix_html/format.ex#L42

Works perfectly, but in my case the text has new lines formatted as `\r\n\r\n`. This pattern is not recognised as double new line, and thus not split into paragraphs:

```
iex> text_to_html("Hello\r\n\r\nWorld") |> safe_to_string
"<p>Hello\r<br>\n\r<br>\nWorld</p>\n"
```

Is there a reason for not including this pattern?

From what I understand `\r\n` (Carriage Return & Line Feed) is used by Windows. 
But my text was written in a `textarea` (Chrome on macOS), and stored and fetched form Postgres (Unix).

---

_Newline characters that major platforms use:_

Platform | Line Break
-- | --
Windows | \r\n (Carriage Return & Line Feed)
Unix/Linux/MacOS X | \n (Line Feed)
MacOS 9 and earlier | \r (Carriage Return)

sn3p:
resolves #190

This PR adds support for `\r\n` (Carriage Return & Line Feed) in [`text_to_html/2`](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Format.html#text_to_html/2).

### DIFF
--- a/lib/phoenix_html/format.ex
+++ b/lib/phoenix_html/format.ex
@@ -39,6 +39,7 @@ defmodule Phoenix.HTML.Format do
 
     string
     |> maybe_html_escape(escape?)
+    |> String.replace("\r", "")
     |> String.split("\n\n", trim: true)
     |> Enum.filter(&not_blank?/1)
     |> Enum.map(&wrap_paragraph(&1, wrapper_tag, attributes, insert_brs?))
--- a/lib/phoenix_html/format.ex
+++ b/lib/phoenix_html/format.ex
@@ -6,9 +6,9 @@ defmodule Phoenix.HTML.Format do
   @doc ~S"""
   Returns text transformed into HTML using simple formatting rules.
 
-  Two or more consecutive newlines `\n\n` are considered as a paragraph
-  and text between them is wrapped in `<p>` tags.
-  One newline `\n` is considered as a linebreak and a `<br>` tag is inserted.
+  Two or more consecutive newlines `\n\n` or `\r\n\r\n` are considered as a
+  paragraph and text between them is wrapped in `<p>` tags.
+  One newline `\n` or `\r\n` is considered as a linebreak and a `<br>` tag is inserted.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/format.ex
```diff
@@ -39,6 +39,7 @@ defmodule Phoenix.HTML.Format do
 
     string
     |> maybe_html_escape(escape?)
+    |> String.replace("\r", "")
```
sn3p: Removing the `\r` is the cleanest way I think.
Could also do add split patterns:

```ex
|> String.split(["\n\n", "\r\n\r\n"], trim: true)
```

But would have to do this in multiple places.

josevalim: It feels semantically wrong though. I would prefer to add split patterns! :+1:

test/phoenix_html/format_test.exs
```diff
@@ -11,8 +11,8 @@ defmodule Phoenix.HTML.FormatTest do
       format("""
       Hello,
 
-      Please come see me.
-
+      Please come see me.\r
```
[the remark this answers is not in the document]
josevalim: :+1: For an extra test instead of adding new ones!

lib/phoenix_html/format.ex
sn3p: > It feels semantically wrong though. I would prefer to add split patterns!

Done! Let me know what you think :)

sn3p: Done! let me know what you think :)

josevalim:

PR #193, opened 2017-12-27
JS escape fixes for uncommon whitespace and null

alvinlindstam:
U+2028 and U+2029 is currently escaped using html escape sequences instead of js escapes, which this PR changes.

It also adds an escape for the U+0000 null character, which is invalid within html script tags.


alvinlindstam:
I have a few other thoughts about this function, that I'd like to discuss first. Is the phoenix-core list appropriate for discussions about phoenix_html or is that just for the phoenix core repo?

josevalim:
> I have a few other thoughts about this function, that I'd like to discuss first. Is the phoenix-core list appropriate for discussions about phoenix_html or is that just for the phoenix core repo?


Yes, please!

josevalim:

PR #195, opened 2018-01-22
Fixing id generation on radio buttons. fixes #194

### LINKED ISSUE #194: `radio_button` id attribute is not generated correctly
See: https://github.com/phoenixframework/phoenix_html/blob/v2.10.5/lib/phoenix_html/form.ex#L668

So if radio button value is "purple elephant", id generated will become `record_field_purple elephant`

I think value needs to be stripped of all spaces before concatenating with `input_id/2` .. or replaced with "_"

GBH:
Made a `input_id/3` just so it's easier to generate `for` attribute on labels. Maybe it makes sense to expand `label` to allow value passing. Not sure. One thing at the time.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -333,6 +333,16 @@ defmodule Phoenix.HTML.Form do
   def input_id(name, field) when is_atom(name),
     do: "#{name}_#{field}"
 
+
+  @doc """
+  Returns an id of a corresponding form field and value attached to it.
+  Useful for radio buttons and inputs like multiselect checkboxes.
+  """
+  def input_id(name, field, value) do
+    value_id = String.replace(value, ~r/\W/u, "_")
+    input_id(name, field) <> "_" <> value_id
+  end
+
   @doc """
   Returns a name of a corresponding form field.
 
@@ -733,7 +743,7 @@ defmodule Phoenix.HTML.Form do
     opts =
       opts
       |> Keyword.put_new(:type, "radio")
-      |> Keyword.put_new(:id, input_id(form, field) <> "_" <> elem(value, 1))
+      |> Keyword.put_new(:id, input_id(form, field, elem(value, 1)))
       |> Keyword.put_new(:name, input_name(form, field))
 
     opts =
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -332,6 +332,10 @@ defmodule Phoenix.HTML.Form do
     do: "#{id}_#{field}"
   def input_id(name, field) when is_atom(name),
     do: "#{name}_#{field}"
+  def input_id(name, field, value) do
```
josevalim: I don't think we need to make this function public. What if we call it `radio_button_id` and keep it private for now?

lib/phoenix_html/form.ex
```diff
@@ -332,6 +332,10 @@ defmodule Phoenix.HTML.Form do
...
+    value_id = value |> String.downcase |> String.replace(~r/\W/, "_")
```
josevalim: Why do we need to downcase?

josevalim: We should probably set the Unicode flag in the regex: `~r/\W/u`.

josevalim:
I have added some comments, thank you @GBH!

lib/phoenix_html/form.ex
GBH: This is basically so I can do `label(name, field, for: input_id(name, field, value))`. Otherwise I gotta re-implement that again. Also this method is useful for checkbox collection (when you have `name: "search[keys][]` for example) and need unique ids just like for radio buttons.

lib/phoenix_html/form.ex
GBH: Downcasing is just to make it look consistent. "search_key_foo" looks better than "search_key_FoO". Gotcha on the unicode thing (coming from ruby I didn't know it's a thing)

lib/phoenix_html/form.ex
josevalim: I see, agreed. If that's the case, we need to add documentation to it, as it is effectively a separate function than `input_id/2`.

lib/phoenix_html/form.ex
josevalim: Let's remove `downcase` then. It is not necessary. If people want it downcase, then they can pass it explicitly using the input_id/3 function. :D

lib/phoenix_html/form.ex
GBH: Will do

lib/phoenix_html/form.ex
GBH: Fair enough. Will update PR for tomorrow.

GBH:
@josevalim PR is amended

Issue #197, opened 2018-02-04
String keys in phoenix_html forms


LostKobrakai:
I'm working on a system, which allows users to build custom forms from building blocks we provide. Now as fields are user defined I'd really like to use string names for those fields and not atoms, but it seems phoenix_html does currently expect atom names for form-fields in various places. 

What's the reasoning for requiring atom keys, when they're converted to strings anyways?

josevalim:
Because we only used atoms so far. ;)

A PR that relax this condition and adds some tests is welcome!

PR #198, opened 2018-02-04
Enable string keys to be usable with forms

LostKobrakai:
Seems like `input_value` / FormData implementations were the only thing really relying on atom keys. The rest were either using `input_value` or doing `Atom.to_string` anyways.

I'd expect this would need `phoenix_ecto` to be updated as well.

Regarding to #197

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -185,11 +185,12 @@ defmodule Phoenix.HTML.Form do
   defstruct source: nil, impl: nil, id: nil, name: nil, data: nil,
             hidden: [], params: %{}, errors: [], options: [], index: nil
 
-  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t, 
-                   data: %{atom => term} | %{String.t => term}, 
-                   params: %{binary => term}, hidden: Keyword.t, options: Keyword.t, 
-                   errors: Keyword.t, impl: module, id: String.t, 
-                   index: nil | non_neg_integer}
+  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t,
+                   data: %{field => term}, params: %{binary => term},
+                   hidden: Keyword.t, options: Keyword.t, errors: Keyword.t,
+                   impl: module, id: String.t, index: nil | non_neg_integer}
+
+  @type field :: atom | String.t
 
   @doc """
   Converts an attribute/form field into its humanize version.
@@ -280,8 +281,8 @@ defmodule Phoenix.HTML.Form do
       sent through the form.
 
   """
-  @spec inputs_for(t, atom | String.t, Keyword.t, (t -> Phoenix.HTML.unsafe)) :: Phoenix.HTML.safe
-  def inputs_for(%{impl: impl} = form, field, options \\ [], fun) do
+  @spec inputs_for(t, field, Keyword.t, (t -> Phoenix.HTML.unsafe)) :: Phoenix.HTML.safe
+  def inputs_for(%{impl: impl} = form, field, options \\ [], fun) when is_atom(field) or is_binary(field) do
     options =
       form.options
       |> Keyword.take([:multipart])
@@ -305,8 +306,8 @@ defmodule Phoenix.HTML.Form do
   fallback to parameters and finally fallback to the default
   struct/map value.
   """
-  @spec input_value(t | atom, atom | String.t) :: term
-  def input_value(%{source: source, impl: impl} = form, field) do
+  @spec input_value(t | atom, field) :: term
+  def input_value(%{source: source, impl: impl} = form, field) when is_atom(field) or is_binary(field) do
     try do
       impl.input_value(source, form, field)
     rescue
@@ -329,12 +330,12 @@ defmodule Phoenix.HTML.Form do
   The form should either be a `Phoenix.HTML.Form` emitted
   by `form_for` or an atom.
   """
-  @spec input_id(t | atom, atom | String.t) :: String.t
+  @spec input_id(t | atom, field) :: String.t
   def input_id(%{id: nil}, _field),
     do: nil
-  def input_id(%{id: id}, field),
+  def input_id(%{id: id}, field) when is_atom(field) or is_binary(field),
     do: "#{id}_#{field}"
-  def input_id(name, field) when is_atom(name),
+  def input_id(name, field) when is_atom(name) and is_atom(field) or is_binary(field),
     do: "#{name}_#{field}"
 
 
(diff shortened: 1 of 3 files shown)


lib/phoenix_html/form.ex
```diff
@@ -185,9 +185,11 @@ defmodule Phoenix.HTML.Form do
   defstruct source: nil, impl: nil, id: nil, name: nil, data: nil,
             hidden: [], params: %{}, errors: [], options: [], index: nil
 
-  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t, data: %{atom => term},
-                   params: %{binary => term}, hidden: Keyword.t, options: Keyword.t,
-                   errors: Keyword.t, impl: module, id: String.t, index: nil | non_neg_integer}
+  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t, 
+                   data: %{atom => term} | %{String.t => term},
```
josevalim: `%{(atom | String.t) => term}` should be preferred as we are effectively allowing mixed keys?

lib/phoenix_html/form.ex
```diff
@@ -303,12 +305,13 @@ defmodule Phoenix.HTML.Form do
...
   """
-  def input_value(%{source: source, impl: impl} = form, field) when is_atom(field) do
+  @spec input_value(t | atom, atom | String.t) :: term
+  def input_value(%{source: source, impl: impl} = form, field) do
     try do
       impl.input_value(source, form, field)
     rescue
       UndefinedFunctionError ->
-        case Map.fetch(form.params, Atom.to_string(field)) do
+        case Map.fetch(form.params, to_string(field)) do
```
josevalim: This change unfortunately allows other kinds of inputs. We should still support only atoms or strings.

lib/phoenix_html/form.ex
LostKobrakai: Yeah, I always feel mixed keys should be avoided, but nothing in the code speaks against it as well.

lib/phoenix_html/form.ex
```diff
@@ -391,7 +397,7 @@ defmodule Phoenix.HTML.Form do
         "password" => :password_input}
 
   """
-  @spec input_type(t, atom) :: atom
+  @spec input_type(t, atom | String.t) :: atom
```
josevalim: We should create a type named `field`, so we don't repeat `atom | String.t` all around the code.

lib/phoenix_html/form.ex
```diff
@@ -402,7 +408,7 @@ defmodule Phoenix.HTML.Form do
       end
 
     if type == :text_input do
-      field = Atom.to_string(field)
+      field = to_string(field)
```
josevalim: Same here. We want to support only atom or strings, not any type that implements to_string.

lib/phoenix_html/form.ex
```diff
@@ -593,7 +599,7 @@ defmodule Phoenix.HTML.Form do
 
   defp time_input_value(other), do: other
 
-  defp generic_input(type, form, field, opts) when is_atom(field) and is_list(opts) do
```
josevalim: We should keep the guard and add or is_binary(field).

lib/phoenix_html/form.ex
```diff
@@ -278,7 +280,7 @@ defmodule Phoenix.HTML.Form do
       sent through the form.
 
   """
-  @spec inputs_for(t, atom, Keyword.t, (t -> Phoenix.HTML.unsafe)) :: Phoenix.HTML.safe
+  @spec inputs_for(t, atom | String.t, Keyword.t, (t -> Phoenix.HTML.unsafe)) :: Phoenix.HTML.safe
   def inputs_for(%{impl: impl} = form, field, options \\ [], fun) do
```
josevalim: Adding guards here would be nice.

lib/phoenix_html/form.ex
```diff
@@ -303,12 +305,13 @@ defmodule Phoenix.HTML.Form do
   fallback to parameters and finally fallback to the default
   struct/map value.
   """
-  def input_value(%{source: source, impl: impl} = form, field) when is_atom(field) do
+  @spec input_value(t | atom, atom | String.t) :: term
+  def input_value(%{source: source, impl: impl} = form, field) do
```
josevalim: Adding guards here would be nice.

lib/phoenix_html/form.ex
```diff
@@ -326,6 +329,7 @@ defmodule Phoenix.HTML.Form do
   The form should either be a `Phoenix.HTML.Form` emitted
   by `form_for` or an atom.
   """
+  @spec input_id(t | atom, atom | String.t) :: String.t
   def input_id(%{id: nil}, _field),
```
josevalim: Adding guards to the `input_` functions would be nice too.

lib/phoenix_html/form.ex
```diff
@@ -185,9 +185,12 @@ defmodule Phoenix.HTML.Form do
   defstruct source: nil, impl: nil, id: nil, name: nil, data: nil,
             hidden: [], params: %{}, errors: [], options: [], index: nil
 
-  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t, data: %{atom => term},
-                   params: %{binary => term}, hidden: Keyword.t, options: Keyword.t,
-                   errors: Keyword.t, impl: module, id: String.t, index: nil | non_neg_integer}
+  @type t :: %Form{source: Phoenix.HTML.FormData.t, name: String.t,
+                   data: %{(atom | String.t) => term}, params: %{binary => term},
```
josevalim: We can use `field => term` here now. :)

lib/phoenix_html/form.ex
josevalim: I would still like to avoid the dynamic dispatch. Let's create a private function called `field_to_string` and make sure all places invoke it.

lib/phoenix_html/form.ex
```diff
@@ -593,7 +600,7 @@ defmodule Phoenix.HTML.Form do
 
   defp time_input_value(other), do: other
 
-  defp generic_input(type, form, field, opts) when is_atom(field) and is_list(opts) do
+  defp generic_input(type, form, field, opts) when is_list(opts) and is_atom(field) or is_binary(field) do
```
josevalim: The precedence here is wrong. You need to add parens around the `or` operands.

josevalim:
Please rebase and we should be good to go! Thank you! :heart:

LostKobrakai:
I'm just wondering, because I had to fix the spec (mostly regarding the other PR). Should `input_id/3` really only work with `String.t | atom`?

josevalim:
> I'm just wondering, because I had to fix the spec (mostly regarding the other PR). Should input_id/3 really only work with String.t | atom?

I would prefer to start strict for now. We can always relax it later (the opposite is not true).

PR #199, opened 2018-02-04
Make id generation of radio buttons work with values, which are html escaped

LostKobrakai:


josevalim:

PR #201, opened 2018-02-22
input_name should return a String when passed an unnamed form.

schneiderderek:
- This bug shows up when using multiple_select with an unnamed form.
- Function spec indicates that input_name returns a String.


josevalim:

PR #203, opened 2018-03-17
Allow value option on labels for easier building of dynamic radio buttons/checkboxes

mintcore:
This will allow labels to be used with dynamically created radiobuttons and checkboxes to have the correct prefix.

For example this will now be allowed:

```elixir
radio_button :user, :level, :gold
label :user, :level, value: :gold
```

which will produce:

```html
<input id="user_level_gold" name="user[level]" type="radio" value="gold">
<label for="user_level_gold"></label>
```

It'll be an improvement over the current way where you'd either have to build the `for` attribute yourself.

Could be a followup on #195

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1433,6 +1433,9 @@ defmodule Phoenix.HTML.Form do
         "E-mail Address"
       end
       #=> <label class="control-label" for="user_email">E-mail Address</label>
+
+      label :user, :newsletters, "Recieve all newsletters", value: "all"
+      #=> <label for="user_newsletters_all">Recieve all newsletters</label>
   """
   def label(form, field) do
     label(form, field, humanize(field), [])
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @mintcore! I see the need for this feature but I don't like using `:value` because maybe something somewhere expects a value attribute in labels and this would make it impossible. We need a way to not clobber the attribute options but I don't have an idea right now.

Thoughts?

mintcore:
It's a good point @josevalim! I've thought about that some too but wanted to hear more opinions on it. 

I'll take some time to think and get back on it when i've got some new ideas!

josevalim:
One option is to introduce this:

```elixir
label :user do 
  radio_button :user, :level, :gold
end
```

Because if the input is inside the label, you don't need `for`.

mintcore:
One way could be to allow `field` to be a tuple

```elixir
label :user, {:level, :gold} do 
  radio_button :user, :level, :gold
end
```

But it does not feel like something anyone would figure out by themselves and therefore not that good.

mintcore:
That is true! While it covers many cases I'm afraid it won't cover them all.

josevalim:
@mintcore for the remaining cases you can always pass `for: input_id(...)`.

mintcore:
@josevalim True! So that basically means we can close this, right?

josevalim:
@mintcore not quite because we don't support label block but I am adding it now. :)

mintcore:
@josevalim Let me know if there is anything more I can do!

josevalim:
Now in master, thanks! :)

PR #205, opened 2018-03-21
Fix inconsistencies with time input.

schneiderderek:
- Allow truncation precision to be configured.
- Get value from input_value to perform truncation in case value was not
specified in opts. <- This fixes the inconsistencies with the input value.
- Run `mix format` (Elixir 1.6.3)

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -629,28 +629,45 @@ defmodule Phoenix.HTML.Form do
   Warning: this feature isn't available in all browsers.
   Check `http://caniuse.com/#feat=input-datetime` for further informations.
 
-  See `text_input/3` for example and docs.
+  ## Options
+
+    * `:precision` - Allowed values: `:minute`, `:second`, `:millisecond`.
+      Defaults to `:minute`.
+
+  All other options are forwarded. See `text_input/3` for example and docs.
+
+  ## Examples
+
+      time_input form, :time
+      #=> <input id="form_time" name="form[time]" type="time" value="23:00">
+
+      time_input form, :time, precision: :second
+      #=> <input id="form_time" name="form[time]" type="time" value="23:00:00">
+
+      time_input form, :time, precision: :millisecond
+      #=> <input id="form_time" name="form[time]" type="time" value="23:00:00.000">
   """
   def time_input(form, field, opts \\ []) do
-    opts =
-      case Keyword.fetch(opts, :value) do
-        {:ok, value} ->
-          Keyword.put(opts, :value, time_input_value(value))
-
-        :error ->
-          opts
-      end
+    {precision, opts} = Keyword.pop(opts, :precision, :minute)
+    value = opts[:value] || input_value(form, field)
+    opts = Keyword.put(opts, :value, truncate_time(value, precision))
 
     generic_input(:time, form, field, opts)
   end
 
-  defp time_input_value(%Time{} = value) do
-    value
-    |> to_string
-    |> String.slice(0..4)
+  defp truncate_time(%Time{} = time, :minute) do
+    time
+    |> Time.to_string()
+    |> String.slice(0, 5)
   end
 
-  defp time_input_value(other), do: other
+  defp truncate_time(%Time{} = time, precision) do
+    time
+    |> Time.truncate(precision)
+    |> Time.to_string()
+  end
+
+  defp truncate_time(value, _), do: value
 
   defp generic_input(type, form, field, opts)
        when is_list(opts) and (is_atom(field) or is_binary(field)) do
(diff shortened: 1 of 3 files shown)


lib/phoenix_html/form.ex
```diff
@@ -629,29 +629,35 @@ defmodule Phoenix.HTML.Form do
   Warning: this feature isn't available in all browsers.
   Check `http://caniuse.com/#feat=input-datetime` for further informations.
 
-  See `text_input/3` for example and docs.
+  ## Options
+
+    * `:precision` - precision passed to `Time.truncate/2`. Can only be used if
+       value is a `%Time{}`.
```
josevalim: Can you please explain the default and include an example?

lib/phoenix_html/form.ex
```diff
@@ -629,29 +629,35 @@ defmodule Phoenix.HTML.Form do
...
+
+  All other options are forwarded. See `text_input/3` for example and docs.
   """
   def time_input(form, field, opts \\ []) do
-    opts =
-      case Keyword.fetch(opts, :value) do
-        {:ok, value} ->
-          Keyword.put(opts, :value, time_input_value(value))
+    value = case Keyword.fetch(opts, :value) do
```
josevalim: `value = opts[:value] || input_value(form, field)`

schneiderderek:
Also I didn't realize that `Time.truncate` was just added in `1.6.0`, so maybe I shouldn't use that.

lib/phoenix_html/form.ex
```diff
@@ -629,29 +629,35 @@ defmodule Phoenix.HTML.Form do
...
 
         :error ->
-          opts
+          input_value(form, field)
+    end
+
+    opts =
+      case value do
+        %Time{} = time ->
+          {precision, opts} = Keyword.pop(opts, :precision, :millisecond)
```
josevalim: We always need to pop the precision option. So we should probably move it up.

test/phoenix_html/form_test.exs
```diff
@@ -449,19 +450,26 @@ defmodule Phoenix.HTML.FormTest do
 
     assert safe_to_string(
              time_input(:search, :key, value: ~T[23:00:07.001], id: "key", name: "search[key][]")
-           ) == ~s(<input id="key" name="search[key][]" type="time" value="23:00">)
+           ) == ~s(<input id="key" name="search[key][]" type="time" value="23:00:07.001">)
```
josevalim: This is changing the behaviour, no? I think our default precision should be `:minute` and we handle it especially.

test/phoenix_html/form_test.exs
```diff
@@ -449,19 +450,26 @@ defmodule Phoenix.HTML.FormTest do
...
     assert safe_form(&time_input(&1, :key)) ==
              ~s(<input id="search_key" name="search[key]" type="time" value="value">)
 
+    assert safe_form(&time_input(&1, :time)) ==
+             ~s(<input id="search_time" name="search[time]" type="time" value="01:02:03.004">)
+
+    assert safe_form(&time_input(&1, :time, precision: :second)) ==
+             ~s(<input id="search_time" name="search[time]" type="time" value="01:02:03">)
+
+
```
josevalim: Please run the formatter on this file as it wouldn't keep double space.

josevalim:
> Also I didn't realize that Time.truncate was just added in 1.6.0, so maybe I shouldn't use that?

Let's default to `:minute` and, if that's the case, we have the same behaviour as before and we don't invoke `truncate`. We will only call truncate if another precision is given.

schneiderderek:
@josevalim I believe i've addressed all of your concerns. Let me know if I should change how i'm handling the `Time.truncate/2` issue.

lib/phoenix_html/form.ex
```diff
@@ -629,28 +629,43 @@ defmodule Phoenix.HTML.Form do
...
-    value
+  defp truncate_time(%Time{} = time, :minute) do
+    time
     |> to_string
-    |> String.slice(0..4)
+    |> String.slice(0, 5)
+  end
+
+  defp truncate_time(%Time{} = time, precision) do
+    Time.truncate(time, precision)
```
josevalim: Same here: `time |> Time.truncate(precision) |> Time.to_string`.

josevalim:
@schneiderderek two minor notes and we should be good to go!

schneiderderek:
@josevalim Fixed, thanks for the feedback! 👍

josevalim:
You will also have to wrap the test that uses precision to run only Elixir v1.6. Something like:

    if Version.match?(System.version, ">= 1.6") do

schneiderderek:
@josevalim Fixed. I wasn't sure how to handle that, thanks! 👍

josevalim:
:heart: :green_heart: :blue_heart: :y...

Issue #207, opened 2018-04-03
Why link helper is replacing the href value to "#" for methods other than GET?


nicolasblanco:
Hi,

when using the `link` helper with a method other than GET, the `href` attribute of the generated `<a>` tag is set to `"#"`, this can be seen there: https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/link.ex#L106

I know this is purely cosmetic but I like to see the full destination address in the browser status bar when hovering on a link. I think this is also the behaviour when generating links with the Rails framework.

Is there a specific reason not to simply always let the full URL in the `href` attribute?

josevalim:
Sounds good to me. PR please?

hlhr:
A few days ago I thought the same, I will send a PR.

wojtekmach:
Fixed by #209.

PR #209, opened 2018-04-16
Replace hash symbol with destination address in link helper

hlhr:
Changes for issuee #207 

Before the change every href attribute got set to `#` in the link helper if the method option isn't `:get`. This PR changes the current behaviour to set the href attribute to the destination address. That shouldn't have any side effects because both link to the current page.


josevalim:

Issue #206, opened 2018-03-22
Feature proposal: data-prompt


begedin:
I like the general approach for the `data-confirm` functionality. Using that as an example, I managed to implement my own version of a `data-prompt` where the behavior is basically

- attributes `data-prompt-message` and `data-prompt-value`
- when the link is clicked, `window.prompt` is called and the user needs to type in the exact value in `data-prompt-value`. If the prompt is dismissed or another value is entered
- both `data-prompt-message` and `data-prompt-value` need to be present for the prompt to trigger

Unfortunately, due to phoenix_html.js enhancing all links generally, I also have to call `stopImmediatePropagation` in my code and basically repeat the entirety of the form creation logic

It works for my specific case, but it got me thinking, would a `window.prompt` be a welcome feature for this library. If so, I could submit a PR first chance I get.

josevalim:
@begedin I would prefer to find a way to make the functionality we have right now easier to compose and/or reuse then adding more feature to the javascript.

begedin:
@josevalim Understandable. Is there any documentation or discussion I could take a look at to understand better what you had in mind? My guess would be a sort of rewrite to separate the `data-confirm` feature from the general link enhancement feature?

josevalim:
@begedin I have no idea as JS is not my strong suite but some sort of callback that would allow you to hook into the form (or similar).

begedin:
If it's acceptable, I can try to work something up (not immediately, need to find the time for it first) and submit a PR as a proposal. At the very least, it can be a starting point for a discussion.

I'm thinking this can probably be closed then.

josevalim:
Beautiful, thank you! :heart:

PR #219, opened 2018-06-01
Configurable and extendable data-confirm behaviour

LostKobrakai:
This change does allow for two things 

* People can customize the data-confirm behaviour:
```javascript
// document.body, so it's executed before the default of phoenix_html
document.body.addEventListener('phoenix.link', function (e) {
  // Prevent default behaviour
  e.stopPropagation();
  
  // Introduce alternative behaviour
  var message = e.target.getAttribute("data-confirm");
  if(!message){ return true; }
  vex.dialog.confirm({
    message: message,
    callback: function (value) {
      if (value == false) { e.preventDefault(); } 
    }
  })
}, false);
```

* Or add their own logic besides the phoenix data-confirm default behaviour (https://github.com/phoenixframework/phoenix_html/issues/206):

```javascript
window.addEventListener('phoenix.link', function (e) {
  // Introduce custom behaviour
  var message = e.target.getAttribute("data-prompt");
  var answer = e.target.getAttribute("data-prompt-answer");
  if(message && answer && (answer != window.prompt(message))) {
    e.preventDefault();
  }
}, false);
```

I've not looked extensively into cross browser support for it by now, but wanted to propose the idea first. There might be some work needed on that front.

[Demo](https://jsfiddle.net/lostkobrakai/54t1a1po/2/)

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -42,21 +42,20 @@ defmodule Phoenix.HTML.Link do
 
   All other options are forwarded to the underlying `<a>` tag.
 
-  ## Data attributes
+  ## JavaScript dependency
 
-  Data attributes are added as a keyword list passed to the
-  `data` key. The following data attributes are supported:
+  In order to support links where `:method` is not `:get` or use the above
+  data attributes, `Phoenix.HTML` relies on JavaScript. You can load
+  `priv/static/phoenix_html.js` into your build tool.
+
+  ### Data attributes
+
+  Data attributes are added as a keyword list passed to the `data` key.
+  The following data attributes are supported:
 
     * `data-confirm` - shows a confirmation prompt before
       generating and submitting the form when `:method`
       is not `:get`.
-
-  ## JavaScript dependency
-
-  In order to support links where `:method` is not `:get`
-  or use the above data attributes, `Phoenix.HTML` relies
-  on JavaScript. You can load `priv/static/phoenix_html.js`
-  into your build tool.
 
   ### Overriding the default confirm behaviour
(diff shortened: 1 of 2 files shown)


josevalim:
@LostKobrakai sounds good to me. I would just name the event "phoenix.click" or "phoenix.link.click". What do you think?

Regarding polyfills: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill

LostKobrakai:
Yeah, adding the action to the event name seems quite reasonable. Though I'm not sure regarding which of your proposed names is better.

Regarding the polyfill:
I don't think we need `CustomEvents`. At least https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events suggests that's only needed when the event needs to hold custom data. But the same article speaks of `new Event('phoenix.click')` not being supported in IE, while the IE supported way is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent). So that's where we need a solid solution.

LostKobrakai:
I just implemented the polyfill of mdn onto current master.

josevalim:
Perfect. I think we just need documentation now. Probably around the same area as this PR: #221.

josevalim:

simonprev:
This seems to have broken the master branch. `link` is not defined anywhere and is used `!link.dispatchEvent....`. I think it should be `!element.dispatchEvent`

chrismccord:
Good catch. Can you send a PR? Thanks!

simonprev:
Done! #224 😄

PR #220, opened 2018-06-20
Support ISO 8601 formatted strings for date values.

schneiderderek:
This change allows ISO 8601 formatted strings to be passed as the value for date inputs.

Specifically, we're have an issue where we're passing an ISO 8601 string to a `date_select`.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1350,6 +1350,13 @@ defmodule Phoenix.HTML.Form do
   defp time_value({hour, min, sec}), do: %{hour: hour, minute: min, second: sec}
 
   defp time_value(nil), do: %{hour: nil, minute: nil, second: nil}
+
+  defp time_value(string) when is_binary(string) do
+    string
+    |> Time.from_iso8601!()
+    |> time_value
+  end
+
   defp time_value(other), do: raise(ArgumentError, "unrecognized time #{inspect(other)}")
 
   @months [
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -1300,6 +1301,13 @@ defmodule Phoenix.HTML.Form do
   defp date_value({year, month, day}), do: %{year: year, month: month, day: day}
 
   defp date_value(nil), do: %{year: nil, month: nil, day: nil}
+
+  defp date_value("" <> string) do
+    string
+    |> Date.from_iso8601!()
+    |> date_value
```
schneiderderek: This `date_value` call isn't required (the tests pass without it), but I figured for consistency it would be a good idea 🤷‍♂️

lib/phoenix_html/form.ex
```diff
@@ -1300,6 +1301,13 @@ defmodule Phoenix.HTML.Form do
   defp date_value({year, month, day}), do: %{year: year, month: month, day: day}
 
   defp date_value(nil), do: %{year: nil, month: nil, day: nil}
+
+  defp date_value("" <> string) do
```
davydog187: `is_binary()`?

schneiderderek: Will change, this is just a shortcut i've used

josevalim:
Can we also support this for Time? Thank you!

schneiderderek:
@josevalim Sure thing 👍

PR #223, opened 2018-06-25
Make input_id/2 return the field name when the form has no id

### LINKED ISSUE #222: input_id/3 argument error when used in a form without :as option
At line 388 in `form.ex`, the `input_id/3` function returns:

```elixir 
input_id(name, field) <> "_" <> value_id
```

This raises an ArgumentError when `input_id(name, field)` returns `nil`, because it cannot be concatenated with the rest of the string.  

Despite the fact that the function specification for `input_id/2` says that it should return a string, it will actually return `nil` when the first argument is a `Phoenix.HTML.Form` that has `id: nil` (line 368):

```elixir 
@spec input_id(t | atom, field) :: String.t()
def input_id(%{id: nil}, _field), do: nil
```

And a form's `id` will be `nil` if the form was created without the `:as` option being set.

Since `input_id/3` is called within `radio_button/4`, this error occurs when trying to add a radio button within a form that does not have the `:as` option set.

This could be fixed either by making sure `input_id/2` always returns a string (as per the spec), or by checking `input_id/2`'s return type when it's called within `input_id/3` and handling accordingly.

jasmineleonard:
Fixes #222 

When `input_id/2` is passed a form with no `id`, it now returns the name of the field rather than `nil`.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -366,7 +366,7 @@ defmodule Phoenix.HTML.Form do
   by `form_for` or an atom.
   """
   @spec input_id(t | atom, field) :: String.t()
-  def input_id(%{id: nil}, _field), do: nil
+  def input_id(%{id: nil}, field), do: "#{field}"
 
   def input_id(%{id: id}, field) when is_atom(field) or is_binary(field) do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #224, opened 2018-07-16
Fix typo 'link' instead of 'element' on click handler

simonprev:
`link` is not defined. The `dispatchEvent` should be called on the event target.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent


chrismccord:
❤️❤️❤️🐥🔥

PR #228, opened 2018-08-23
add errors in the conn to the form

tarzan:
When creating a form with the conn instead of with a changeset struct, there is no easy way of displaying error messages.

This is an attempt to make that easier by allowing `:errors` as an option to `form_for/4`.
In the controller:
```ex
def my_post_method(conn, params) do
  case do_stuff(params) do
    {:ok, object} -> render(conn, view_template)
    {:error, reason} -> render(conn, view_template, errors: reason)
  end
end
```
and in the template
```eex
<%= form_for @conn, "/do_stuff", [errors: @conn.assigns[:errors]], fn f -> do %>
  <%= text_input f, :field %>
  <%= error_tag f, :field %>
<% end %>
```
where `error_tag` is the custom function as mentioned in the [migration guides](https://gist.github.com/chrismccord/557ef22e2a03a7992624).

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -93,6 +93,24 @@ defmodule Phoenix.HTML.Form do
   request parameters. In this case, the input's value would be set
   to `@conn.params["search"]["for"]`.
 
+  #### A note on `:errors`
+
+  When you feed `form_for/4` with `:errors` as an option you will be able to use
+  your favorite `error_tag`-helper like:
+
+    def error_tag(form, field) do
+      if error = form.errors[field] do
+        content_tag :span, translate_error(error), class: "help-block"
+      end
+    end
+
+  For example, put the errors in the conn.assigns and create your form like this:
+
+    <%= form_for @conn, '/', [errors: @conn.assigns[:errors]], fn f -> %>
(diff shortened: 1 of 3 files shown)


chrismccord:
I think in this case it's better to handle the errors yourself, for example you could simply use the `@errors` assign yourself in the template rather than lifting it off the form.

josevalim:
The advantage of lifting of the form is that the error tag helper that we have just works but i am not sure how comfortable i am with hijacking the assigns errors key. Maybe we should rather pass errors as an explicit option on `form_for`?

josevalim:
By passing an option maybe we get the best of both worlds. Unsure.

chrismccord:
Ah good point. I handle thought about error_tag. I really like the idea of passing the errors as an option to form_for! That would get my vote.

tarzan:
Thanks for the swift response(s)! I'll rework this as an option for `form_for`.

tarzan:
Hey @chrismccord and @josevalim I changed it to allowing `:errors` as an option to `form_for/4`. 

Added a bit of documentation as well. Hope you like it.

Cheers!

josevalim:

Issue #232, opened 2018-10-25
Option to disable unchecked_value for checkbox


LostKobrakai:
Sometimes (e.g. when using a filter form and GET requests) one does not need to send unchecked values, but just what actually has been selected. It would be nice if there would be an option for that in the checkbox function.

LostKobrakai:
Maybe something like `checkbox f, :test, unchecked_input: :skip`. I'm actually not to keen on adding another function, as it's still a checkbox.

josevalim:
Maybe `hidden_input: false`?

LostKobrakai:
Would work for me.

josevalim:
PR please? :)

LostKobrakai:
I'll see what I can do. I'll be currently quite busy then next days and was just leaving this here. Maybe something for hacktoberfest?

PR #233, opened 2018-10-28
Add option to not send values from unchecked checkbox

feliperenan:
fixes https://github.com/phoenixframework/phoenix_html/issues/232

Sometimes we don't want to send values from unchecked checkboxes to the server (e.g. when using a filter form and GET requests).

This PR allows developers to set the option `hidden_input: false` to not generate the hidden input. This way, the value will not be sent to the server if the checkbox is unchecked.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -877,6 +877,9 @@ defmodule Phoenix.HTML.Form do
     * `:checked_value` - the value to be sent when the checkbox is checked.
       Defaults to "true"
 
+    * `:hidden_input` - the value to check if a checkox will generate a hidden input or not,
+      Defaults to "true"
+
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -877,8 +877,8 @@ defmodule Phoenix.HTML.Form do
     * `:checked_value` - the value to be sent when the checkbox is checked.
       Defaults to "true"
 
-    * `:hidden_input` - the value to check if a checkox will generate a hidden input or not,
-      Defaults to "true"
+    * `:hidden_input` - the value to check if a checkox will generate a hidden input
```
feliperenan: `checkox` => `checkbox` 🙈

josevalim:

PR #234, opened 2018-10-31
Optimize documentation for `Form.HTML.input_name/2`

mschae:
Super small change but I tripped over the documentation a few times since the first param is parsed and displayed as `name` although it can be `form` or `name`. The documentation says as much but then again, who reads documentations right? 😅


josevalim:

PR #240, opened 2018-12-13
remove match that cannot succeed

lukaszsamson:
fixes dialyzer warning
Guard test __@1::{_,[{atom(),_}]} =:= 'false' can never succeed


josevalim:

PR #243, opened 2019-01-03
Implement ability to pass in HTML safe strings to label/4

alizain:
A couple of the other HTML functions allow passing in HTML safe tuples directly, in lieu of text. This PR adds the same functionality for the `label/4` function.

With this, you can do something like:

```elixir
<%= label @applicant_form, :additional_info, {:safe, "Additional <span>Information</span>"} %>
```

On the intent itself - I can agree that passing in HTML as a `:safe` tuple directly is not good idea. As mentioned, I've often overloaded the `text` arg in `link(text, opts)` to allow HTML (not realizing that the `link/2` function allows `do` blocks as well.

The only good reason for merging this is in because we usually passthrough the `text` argument eventually to `html_escape`, which is responsible for making this value usable. It seems arbitrary and inconsistent for some APIs to passthrough without prejudice (such as the `link/2` tag), and for other APIs to have the `when is_binary(text)` guard clause.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1615,10 +1615,6 @@ defmodule Phoenix.HTML.Form do
   """
   def label(form, field, text_or_do_block_or_attributes)
 
-  def label(form, field, text) when is_binary(text) do
-    label(form, field, text, [])
-  end
-
   def label(form, field, do: block) do
     label(form, field, [], do: block)
   end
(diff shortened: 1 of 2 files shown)


josevalim:
@alizain with a label you would achieve that by using do/end blocks:

```elixir
<%= label @applcant_form, :additional_info do
  {:safe, "Additional <span>Information</span>"} 
end %> 
```

What are the other HTML functions that allow the safe tuples directly. To be clear, I *think* we should merge this, but I want to make sure we are consistent. :)

alizain:
Yeah I saw that, it's what I'm going to start using everywhere now.

I really like that do/end blocks are predictably available in most of the APIs I've used in Phoenix.HTML thus far.

As mentioned in my edited comment, `link/2` allows safe tuples directly... not sure which other ones do. I wouldn't mind adding tests for each case though, to figure out which ones do and which ones don't

josevalim:
@alizain thanks. I do think we allow the safe content in most cases but we didn't add it to `label` because `label` already has a "diversified" API. Anyway, this is good go to, thanks!

josevalim:

alizain:
Thank you so much!

PR #244, opened 2019-01-04
[FIX] deprecation warning referred to deprecated function

nduitz:
As mention in commit deprecation warning referes to the deprecated function itself


josevalim:

PR #255, opened 2019-04-23
Fix typespec for form_for/2-3

### LINKED ISSUE #239: form_for dialyzer errors after update to 2.13
Version
Erlang/OTP 21 [erts-10.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]
Elixir 1.7.4 (compiled with Erlang/OTP 21)
phoenix 1.4.0
phoenix-html 2.13.0

Current behavior
The code
```
<%= form_for @changeset, @action, fn f -> %>
...
```
results in following dialyzer error
```
Function 'my_form.html'/1 has no local return
The call 'Elixir.Phoenix.HTML.Form':form_for(any(),any(),fun((_) -> {'safe',[any(),...]})) will never return since it differs in the 3rd argument from the success typing arguments: (any(),any(),maybe_improper_list())
```

Expected behavior
dialyzer should not complain as in earlier versions

wojtekmach:
Closes #239

One small problem with current approach was that docs showed both spec clauses in docs for form_for/3 which was supposed to be only about the variant without anonymous function. See https://github.com/phoenixframework/phoenix_html/commit/cae0f75e51fe53de8edb783173790f235cb74604 for more details. When I removed the clause, however, dialyzer complains again, so it needs to be there.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -378,8 +378,6 @@ defmodule Phoenix.HTML.Form do
   See `Phoenix.HTML.Tag.form_tag/2` for more information on the
   options above.
   """
-  @spec form_for(Phoenix.HTML.FormData.t(), String.t(), (t -> Phoenix.HTML.unsafe())) ::
-          Phoenix.HTML.safe()
   @spec form_for(Phoenix.HTML.FormData.t(), String.t(), Keyword.t(), (t -> Phoenix.HTML.unsafe())) ::
           Phoenix.HTML.safe()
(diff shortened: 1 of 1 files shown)


josevalim:

joaquinalcerro:
Hi, I was updating my phoenix_html dependency to fix this issue but I am not getting the updated/fixed code. I can see it was fixed on the master branch but the tag v2.13.2 does not have the updated code.
When I updated the dependency, it updated to the v2.13.2.
Are you going to push a new version to hex?
Thanks for the help.

PR #256, opened 2019-04-23
Use application inference in mix.exs

wojtekmach:


josevalim:

michalmuskala:
Does phoenix_html really need eex at runtime?

wojtekmach:
I thought so, it's used e.g. here: https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/engine.ex#L73 - if this is a compile-time concern only though I'm happy to change it back.

(I run into this issue when running dialyxir, it complained it couldn't find eex.)

michalmuskala:
My reasoning for not including it would be that most people wouldn't actually need it and for those applications that would use the engine at runtime they have to use it as an option to eex itself, so it already has to include a runtime dependency to eex.

josevalim:
EEx is so small that I would rather include it for the 1% of users who may not need it rather than not include it and have folks notice eex is missing when they assemble a release.

michalmuskala:
My point was that you can't use the parts of phoenix_html at runtime without explicitly using eex at runtime directly. 

I agree that it is small, but by that reasoning any single package is small.

Issue #258, opened 2019-04-25
`inputs_for` option to not automatically generate the id field


LostKobrakai:
Currently a structure like the following is not possible using `inputs_for` because the hidden id input is always added as sibling to the user defined markup. Maybe the markup for the id input could optionally be supplied to the callback, so the user can place it somewhere.

```eex
<ol>
  <%= inputs_for f, :fields, fn finner -> %>
    <li>[…]</li>
  <% end %>
</ol>
```

```html
<ol>
  <input type="hidden" …>
  <li>[…]</li>

  <input type="hidden" …>
  <li>[…]</li>

  <input type="hidden" …>
  <li>[…]</li>
</ol>
```

josevalim:
A PR to add an option that skips the id field (so you add it manually) is welcome!

PR #259, opened 2019-04-25
Allow `inputs_for` to skip hidden fields for more control over the final markup

LostKobrakai:
For #258

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -412,7 +412,7 @@ defmodule Phoenix.HTML.Form do
 
     * `:skip_hidden` - skip the automatic rendering of hidden
       fields to allow for more tight control over the generated
-      markup. You can use `hidden_fields_of_form(form)` to 
+      markup. You can use `hidden_form_inputs(form)` to
       generate them manually within the supplied callback.
 
   """
@@ -420,7 +420,7 @@ defmodule Phoenix.HTML.Form do
   def inputs_for(%{impl: impl} = form, field, options \\ [], fun)
       when is_atom(field) or is_binary(field) do
     {skip, options} = Keyword.pop(options, :skip_hidden, false)
-    
+
     options =
       form.options
       |> Keyword.take([:multipart])
@@ -433,16 +433,16 @@ defmodule Phoenix.HTML.Form do
         if skip do
           fun.(form)
         else
-          [hidden_fields_of_form(form), fun.(form)]
+          [hidden_form_inputs(form), fun.(form)]
         end
       end)
     )
   end
-  
+
   @doc false
-  @spec hidden_fields_of_form(t) :: [Phoenix.HTML.safe()]
-  def hidden_fields_of_form(%{hidden: hidden}) do
-    Enum.map(hidden, fn {k, v} -> hidden_input(form, k, value: v) end)
+  @spec hidden_form_inputs(t) :: [Phoenix.HTML.safe()]
+  def hidden_form_inputs(form) do
+    Enum.map(form.hidden, fn {k, v} -> hidden_input(form, k, value: v) end)
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/phoenix_html/form.ex
```diff
@@ -423,11 +430,20 @@ defmodule Phoenix.HTML.Form do
...
+        else
+          [hidden_fields_of_form(form), fun.(form)]
+        end
       end)
     )
   end
+  
+  @doc false
+  @spec hidden_fields_of_form(t) :: [Phoenix.HTML.safe()]
+  def hidden_fields_of_form(%{hidden: hidden}) do
```
josevalim: Shall we call it `hidden_form_inputs/1`?

LostKobrakai:
I've renamed the helper. For testing the new functionality we would need a new implementation of `Phoenix.HTML.FormData` though. `Plug.Conn` doesn't use any hidden fields.

josevalim:
Let’s do that then! We can pass the hidden fields as an option to this new
implementation.
-- 


lib/phoenix_html/form.ex
```diff
@@ -410,10 +410,17 @@ defmodule Phoenix.HTML.Form do
       applies if the field value is a list and no parameters were
       sent through the form.
 
+    * `:skip_hidden` - skip the automatic rendering of hidden
```
feliperenan: We could call this option of `skip_hidden_inputs` to be more explicit. Wdyt?

feliperenan:
Could we only add the `hidden` option in the `Phoenix.HTML.FormData`? I mean something like that:

```Elixir
def to_form(conn_or_atom, form, field, opts) when is_atom(field) or is_binary(field) do
  #...
  {hidden, opts} = Keyword.pop(opts, :hidden, [])
  #...

  cond do
    # cardinality: one
    is_map(default) ->
      [
        %Phoenix.HTML.Form{
          source: conn_or_atom,
          impl: __MODULE__,
          id: id,
          name: name,
          data: default,
          params: params || %{},
          hidden: hidden,
          options: opts
        }
      ]

    #...
  end
  #...
end
```

https://github.com/phoenixframework/phoenix_html/blob/837fdbf2712bb2f38dcd3dc0e828847a2d49a39b/lib/phoenix_html/form_data.ex#L88

I know that we would be doing that only because the test, but would be easier than creating a new implemention (I guess).

If so, we could add these tests :).

```Elixir
describe "inputs_for/4" do
  test "generate a new form builder for the given parameter" do
    conn = conn()

    form =
        form_for(conn, "/", [as: :user], fn form ->
          inputs_for(form, :company, fn company_form ->
            text_input company_form, :name
          end)
        end)
        |> safe_to_string()

    assert form =~ ~s(<input name="_utf8" type="hidden" value="‚úì">)
    assert form =~ ~s(<input id="user_company_name" name="user[company][name]" type="text">)
  end

  test "generate a new form builder with hidden id fields when there are present" do
    conn = conn()

    form =
        form_for(conn, "/", [as: :user], fn form ->
          inputs_for(form, :company, [hidden: [id: 1]],fn company_form ->
            text_input company_form, :name
          end)
        end)
        |> safe_to_string()

    assert form =~ ~s(input id="user_company_id" name="user[company][id]" type="hidden" value="1">)
    assert form =~ ~s(<input id="user_company_name" name="user[company][name]" type="text">)
  end

  test "skip hidden inputs" do
    conn = conn()

    form =
        form_for(conn, "/", [as: :user], fn form ->
          inputs_for(form, :company, [skip_hidden: true, hidden: [id: 1]],fn company_form ->
            text_input company_form, :name
          end)
        end)
        |> safe_to_string()

    refute form =~ ~s(input id="user_company_id" name="user[company][id]" type="hidden" value="1">)
    assert form =~ ~s(<input id="user_company_name" name="user[company][name]" type="text">)
  end
end
```

@LostKobrakai are you willing to continue working on this? Otherwise I can open another pull-request to finish this work :).

LostKobrakai:
@feliperenan go ahead if you like.

feliperenan:
I've opened another one https://github.com/phoenixframework/phoenix_html/pull/259 using your commit as reference, we could close this one now. 😅

PR #261, opened 2019-06-06
updated datetime_local_input to format default form value if if no value is explicitly set

guiclerc:
Fix an issue with datetime_local_input, the date_time does not get formatted if the value is not explicitly set.


josevalim:

PR #263, opened 2019-06-26
Allow `inputs_for` to skip hidden fields for more control over the final markup

### LINKED ISSUE #258: `inputs_for` option to not automatically generate the id field

feliperenan:
Fixes #258

Continue the @LostKobrakai work https://github.com/phoenixframework/phoenix_html/pull/259

### DIFF
--- a/lib/phoenix_html/form_data.ex
+++ b/lib/phoenix_html/form_data.ex
@@ -91,6 +91,7 @@ defimpl Phoenix.HTML.FormData, for: [Plug.Conn, Atom] do
     {append, opts} = Keyword.pop(opts, :append, [])
     {name, opts} = Keyword.pop(opts, :as)
     {id, opts} = Keyword.pop(opts, :id)
+    {hidden, opts} = Keyword.pop(opts, :hidden, [])
 
     id = to_string(id || form.id <> "_#{field}")
     name = to_string(name || form.name <> "[#{field}]")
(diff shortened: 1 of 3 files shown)


lib/phoenix_html/form_data.ex
```diff
@@ -91,6 +91,7 @@ defimpl Phoenix.HTML.FormData, for: [Plug.Conn, Atom] do
     {append, opts} = Keyword.pop(opts, :append, [])
     {name, opts} = Keyword.pop(opts, :as)
     {id, opts} = Keyword.pop(opts, :id)
+    {hidden, opts} = Keyword.pop(opts, :hidden, [])
```
feliperenan: @josevalim I'm just passing this as an option rather than creating a new implementation, not sure if that's a good idea though 🙈. Could you review that? if it doesn't, I will try to figure out how to creating a new implementation of `FormData` to use only in the tests.

josevalim:

nicklayb:
It doesn't seem deployed yet, any estimates for this to be published on hex?

PR #264, opened 2019-07-26
Fix img_tag/2 examples

jayjun:
1. `Routes.static_path/1` should be `Endpoint.static_path/1`.
2. `static_path` only accepts paths with a leading `/`.
3. Example output better matches real output in latest Phoenix.

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -299,23 +299,17 @@ defmodule Phoenix.HTML.Tag do
   ## Examples
 
       img_tag(user.photo_path)
-      <img src="photo.png">
+      <img src="/photo.png">
 
       img_tag(user.photo, class: "image")
-      <img src="smile.png" class="image">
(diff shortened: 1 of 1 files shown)


josevalim:
> Routes.static_path/1 should be Endpoint.static_path/1

Why? :)

jayjun:
Because `Routes.static_path/1` does not exist? Only `Routes.static_path/2` does, as far as I can tell.

josevalim:
@jayjun ah, yes. Let's use `Routes.static_path(conn, "/...")` then. Thanks, good catch!

jayjun:
Cool, no worries. Just to clarify, you want me to remove `Endpoint.static_path/1`?

josevalim:
Yes, Routes.static_path/2 is preferred.

jayjun:
Done! Travis is failing for unrelated `wget` error for _Elixir: 1.5 OTP Release: 18.3_ only.

josevalim:

PR #266, opened 2019-08-07
Adding mail_to helper

iloveitaly:
Hello! Thanks for all of the great work here.

Before putting more work into this, I wanted to check and see if this is the sort of addition that would be welcome to the `phoenix_html` package. It seems like this package is relatively lightweight compared to the helpers available in rails, and wanted to see if that was intentional or if contributions to add additional helpers are welcome.

Thanks!

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -163,6 +163,11 @@ defmodule Phoenix.HTML.Link do
     link(contents, opts)
   end
 
+  def mail_to(email_address), do: mail_to(email_address, email_address)
+  def mail_to(email_address, name) do
+    link(name, to: "mailto:#{email_address}")
+  end
+
   @doc """
   Generates a button that uses a regular HTML form to submit to the given URL.
 


josevalim:
Thanks, it actually looks good to me!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Hi @iloveitaly! Sorry, this has gone stale but, given it is a very short wrapper around `link`, I think we will be fine without it. Thank you!

iloveitaly:
@josevalim the rails `mail_to` is a lot more robust (URL encodes the email, allows for additional params to customize the message). Are you thinking that a `mail_to` wrapper with that advanced functionality makes sense, but not the simple implementation that is contained in this PR? Or are you trying to keep the surface area of the html helpers small intentionally?

josevalim:
I think it is ok to add mail_to if we provide more functionality around it.

PR #268, opened 2019-08-18
Remove :enforce_utf8 option

mikl:
The hidden `_utf8=✓` form field hasn't been needed since IE 8, and given that browser is now three years in the grave, it might be time to remove this.

Rails, where this concept came from, disabled this last year: https://github.com/rails/rails/pull/32125

This MR does not remove `accept-charset="UTF-8"`. That might also be superfluous, but I haven't researched how well browsers handle that.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -352,11 +352,6 @@ defmodule Phoenix.HTML.Form do
       include an input tag with name `_csrf_token`. When set to false, this
       is disabled
 
-    * `:enforce_utf8` - the form will have an `accept-charset` attribute set
-      to "UTF-8" and a hidden input named `_utf8` containing a unicode
-      character to force the browser to use UTF-8 as the charset. When set to
-      false, this is disabled.
-
     * `:errors` - use this to manually pass a keyword list of errors to the form
       (for example from `conn.assigns[:errors]`). This option is only used when a
       connection is used as the form source and it will make the errors available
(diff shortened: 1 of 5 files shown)


mikl:
The test *are* passing, it's just that the "Elixir: 1.5 OTP Release: 18.3" environment can't be installed, since it's trying to install Erlang from the URL `https://storage.googleapis.com/travis-ci-language-archives/erlang/binaries/ubuntu/16.04/x86_64/erlang-18.3-nonroot.tar.bz2`, that is no longer valid - it gives a 404 error.

josevalim:
Good catch! Although I still think it would be best to set the accept
charset attribute, what do you think?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

mikl:
According to the documentation, it could be left out:

> `accept-charset`: […] character encodings that the server accepts. The browser uses them in the order in which they are listed. The *default value*, the reserved string "UNKNOWN", indicates the *same encoding as that of the document containing the form element*.

(from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form, emphasis mine)

Since Phoenix always delivers the document as UTF-8, we should then get form data back in the same encoding. I can't find any indication that there are any current browsers that do not support this.

josevalim:
So I think we can remove the enforce_utf8 option altogether. People can set
the charset if they want something else.
-- 


mikl:
Alright, I'll update the PR accordingly.

mikl:
Alright, it should be good to go.

josevalim:

PR #269, opened 2019-08-18
Disable testing on OTP 18

mikl:
As also noted [over here](https://github.com/phoenixframework/phoenix_html/pull/268#issuecomment-522332831), installing OTP 18.3 on Travis no longer works due to the release no longer being available at the URL Travis tries to download it from.

Given how relatively ancient OTP 18 is, I guess it might make sense to simply stop supporting it.


josevalim:

PR #270, opened 2019-09-11
`class_list` attributes for HTML tags

alizain:
**Before**

```eex
<%= input @form, @field, class: Enum.join(" ", ["form-control"] ++ bootstrap_error_classes(@form, @field)) %>
```

**After**

```eex
<%= input @form, @field, class_list: ["form-control"] ++ bootstrap_error_classes(@form, @field) %>
```

This PR tries to make this pattern less verbose, and probably more efficient as well. Most people won't know that `Enum.join` isn't always the best idea for performance - passing in `IOData` is better.

The implementation is flexible enough that we could allow other HTML attributes to be set this way, and the string to interleave between members can be changed per attribute. In the case of `class_list`, that string is `" "`. For `style_list`, for example, it could be `"; "`

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -131,8 +131,8 @@ defmodule Phoenix.HTML.Tag do
   end
 
   defp build_attrs(tag, [{k, v} | t], acc) when k in @list_attrs and is_list(v) do
-    {k_actual, interleave_char} = Map.fetch!(@list_attrs_map, k)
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @alizain but we never introduce "pseudo attributes" such as `class_list` because if one day the browser decides to introduce a class-list attribute or there is a library that uses such, then users will never be able to set it. Sorry!

alizain:
That's a good point!

The alternative of "interleaving" spaces when a list is passed into the `class` attribute could be as bad - it could easily interfere with someone legitimately passing in an `IOData` list.

I'll do something else on my side. Thanks for the response!

PR #274, opened 2019-10-21
use atom instead of string

sushant12:
https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/link.ex#L206

it is checking for atom.


josevalim:

PR #282, opened 2019-12-13
Allow strings in content_tag

### LINKED ISSUE #281: content_tag/3 doesn't accept a string argument
I'm using the `tag` and `content_tag` functions to generate custom web component elements. Per the spec, custom elements must have dashed names like:

    <my-component></my-component>

The `tag/2` function currently allows a string argument

    iex> safe_to_string tag("my-component")
    "<my-component>"

However, the `content_tag/3` function does not:

    iex> safe_to_string content_tag("my-component", "content")
    ** (FunctionClauseError) no function clause matching in Phoenix.HTML.Tag.content_tag/2

Would it be possible to allow a string argument to the `content_tag` function? 

Another alternative would be to require atoms, but replace underscores with dashes:

    iex> safe_to_string content_tag(:my_component, "content")
    "<my-component>content</my-component>"

That would match the behavior that currently exists for `data` attributes:

    iex> safe_to_string tag(:div, data: [my_attr: "foo"])
    <div data-my-attr="foo">

rockwood:
Closes #281 

This removes the `is_atom` guard clauses on the `content_tag` functions allowing strings to be passed.

    content_tag("custom-component", "Hello")

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -72,19 +72,19 @@ defmodule Phoenix.HTML.Tag do
       "<option data-foo=\"bar\" value=\"value\">Display Value</option>"
 
   """
-  def content_tag(name, do: block) when is_atom(name) do
+  def content_tag(name, do: block) do
     content_tag(name, block, [])
   end
 
-  def content_tag(name, content) when is_atom(name) do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #285, opened 2020-01-28
bugfix(phoenix_html.js): Make IIFE idempotent by injecting a global _…

### LINKED ISSUE #284: Javascript data-confirm fires multiple times
If you're using something like turbolinks, `phoenix_html` may get loaded multiple times. This leads to a situation where users will have to click the confirm dialog over and over to get the dialog to disappear.

Looking at the [code](https://github.com/phoenixframework/phoenix_html/blob/master/priv/static/phoenix_html.js) and the docs for [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), I believe there should be a simple fix to this problem.

> If multiple identical EventListeners are registered on the same EventTarget with the same parameters, the duplicate instances are discarded. They do not cause the EventListener to be called twice, and they do not need to be removed manually with the removeEventListener() method.

> Note, however that when using an anonymous function as the handler, such listeners will NOT be identical, because anonymous functions are not identical even if defined using the SAME unchanging source-code simply called repeatedly, even if in a loop.

It looks like the root of the problem is the fact that we are using anonymous functions when adding the event listener. If we can ensure that the functions are static when the code runs again the additional event listener will be ignored since it is identical.  Since this is based off behavior from rails I think it would be good to borrow/steal their implementation since it should handle these problems out of the box.

https://github.com/rails/rails/tree/master/actionview/app/assets/javascripts

tomciopp:
…phx_loaded boolean flag that is set after setup. Prevents event handlers from being set multiple times in case of libraries like turbolinks which would run this code several times.

@josevalim @chrismccord fixes #284.

After taking some time to think about the problem, I took a different approach than what I outlined in the previous discussion. I think this is a simpler and more robust solution to the problem since there is not much javascript actually being executed. Let me know what you think.

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -1,6 +1,8 @@
 "use strict";
 
-(function() {
+(function(loaded) {
+  if (loaded) { return }
+
   var PolyfillEvent = eventConstructor();
 
   function eventConstructor() {
@@ -73,4 +75,6 @@
       e.preventDefault();
     }
   }, false);
-})();
+
+  window._phx_loaded = true;
+})(window._phx_loaded);


josevalim:
Thanks for the PR!

I am not sure but wouldn’t it be more idiomatic to use a named function? It feels like adding this to work around a limitation on Turbolinks, which is arguably the one in the wrong here, is not very clean.

josevalim:
Closing as stale. :)

PR #290, opened 2020-02-11
Disable hidden input for disabled checkboxes

al2o3cr:
When a checkbox is disabled, it doesn't send its value when the form is submitted - but the still-active hidden input DOES. The result is that a disabled checkbox will always submit as its unchecked value (`"false"` by default).

See also https://github.com/rails/rails/issues/1953 where a similar helper function had the same bug with the same fix.

Not 100% certain about the formatting on the change in `form.ex`; that's what running `mix format` gave me but there's a bunch of other files that that wants to change as well...


josevalim:

Issue #291, opened 2020-02-26
Possibility to use live_component inside of inputs_for


feliperenan:
If we try to use a [live_component](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveComponent.html#content) from Phoenix Live View inside of `inputs_for`, we are going to receive an error like this:

```
** (ArgumentError) cannot convert component SomeComponent to HTML.
```

As far as I understood, it's one of [live_component limitations](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveComponent.html#module-limitations). 

I think we could introduce a similar function like `form_for/3` to `inputs_for` so that this function will not take a function as argument allowing us to render `live_component` inside of it. Something like this:

```EEX
<%= for i <- inputs_for(f, :children) do %>
  // TODO: need to figure out how to generate hidden ids autommatically.
  <div class="field">
    <%= input i, :name %>
  </div>

  <div class="field">
    <%= live_component @socket, SomeComponent, id: :id, form: i %>
  </div>
<% end %>
```

Does it make sense @josevalim? I have similar version working on an App that I've been working and I'm wondering if it would make sense to have it here as well :thinking: .

josevalim:
Yes, it makes total sense. Please send a PR to add such function!

PR #292, opened 2020-02-27
Add Phoenix.HTML.Form.inputs_for/3

feliperenan:
Fixes https://github.com/phoenixframework/phoenix_html/issues/291

TODO: 

- [x] Add docs and specs
- [x] Tests

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -386,10 +386,15 @@ defmodule Phoenix.HTML.Form do
   that replaces the anonymous function for returning the generated forms
   instead.
 
+  Keep in mind that this function does not generate hidden inputs automatically
+  like `inputs/4`. To generate them you need to explicit do it by yourself.
+
       <%= f = form_for @changeset, Routes.user_path(@conn, :create), opts %>
         Name: <%= text_input f, :name %>
 
         <%= for friend_form = inputs_for(f, :friends) do %>
+          # for generating hidden inputs.
+          <%= hidden_inputs_for(company_form) %>
           <%= text_input friend_form, :name %>
         <% end %>
       </form>
@@ -453,8 +458,7 @@ defmodule Phoenix.HTML.Form do
         if skip do
           fun.(form)
         else
-          hidden = Enum.map(form.hidden, fn {k, v} -> hidden_input(form, k, value: v) end)
-          [hidden, fun.(form)]
+          [hidden_inputs_for(form), fun.(form)]
         end
       end)
     )
@@ -631,6 +635,16 @@ defmodule Phoenix.HTML.Form do
   """
   def hidden_input(form, field, opts \\ []) do
     generic_input(:hidden, form, field, opts)
+  end
+
+  @doc """
+  Generates hidden inputs for the given form.
+  """
+  @spec hidden_inputs_for(t) :: list(Phoenix.HTML.safe())
+  def hidden_inputs_for(form) do
+    for {k, v} <- form.hidden do
+      hidden_input(form, k, value: v)
+    end
   end
 
   @doc """
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -373,6 +373,41 @@ defmodule Phoenix.HTML.Form do
   end
 
   @doc """
+  Same as `inputs_for(form, field, [])`.
+  """
+  @spec inputs_for(t, field) :: list(Phoenix.HTML.Form.t())
+  def inputs_for(form, field) when is_atom(field), do: inputs_for(form, field, [])
+
+  @doc """
+  Generate a new form builder for the given parameter in form **without** an
+  anonymous function.
+
+  This functionality exists mostly for integration with `Phoenix.LiveView`
+  that replaces the anonymous function for returning the generated forms
+  instead.
+
+      <%= f = form_for @changeset, Routes.user_path(@conn, :create), opts %>
+        Name: <%= text_input f, :name %>
+
+        <%= for friend_form = inputs_for(f, :friends) do %>
+          <%= text_input friend_form, :name %>
+        <% end %>
+      </form>
+
+  See `inputs_for/4` for the available options.
+  """
+  @spec inputs_for(t, field, Keyword.t()) :: list(Phoenix.HTML.Form.t())
+  def inputs_for(%{impl: impl} = form, field, options)
+      when is_atom(field) and is_list(options) do
+    options =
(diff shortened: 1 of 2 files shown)


feliperenan:
This is still under development. I opened the PR because I have one question regarding hidden inputs. 

I'm not sure how to proceed here. The way I have implemented, `inputs_for/3` will not generate `hidden_inputs` automatically as `inputs_for/4` does. Developers will have to generate them on their own:

```Elixir
for {k, v} <- company_form.hidden do
  hidden_input(company_form, k, value: v)
end
```

I've two suggestions to mitigate that:

1) Accept this trade-off and leave for developers to generate them. We could also provide a public function for that, something like:

```eex
<%= for f <- inputs_for(f, :children) do %>
  <%= hidden_inputs(f) %>

  <%= input f, :name %>
<% end %>
```

2) Return the generated `hidden_fields` along with the form in a tuple format:

```eex
<%- {forms, hidden_inputs} - inputs_for(f, :children) %>
<%= for f <- forms do %>
  <%= hidden_inputs %>

  <%= input f, :name %>
<% end %>
```

Could you help me with that @josevalim 🙌 ?

josevalim:
I think one of the things you need to have in mind is that someone can call `inputs_for` for a single association. In this case, if you do:

```elixir
<%= i = inputs_for(f, :child) do %>
```

I *think* the above will output a form. So first of all, we need a way to say that the no form should be outputted in this case. About the hidden inputs part though, I agree with you, I think it should be a separate function that users will invoke.

feliperenan:
I'm not sure if I got this:

> I think the above will output a form.

The way it is, if you call it like you said:

```eex
<%= i = inputs_for(f, :child) %>
```

You will get the following error:

```
** (ArgumentError) lists in Phoenix.HTML and templates may only contain integers representing bytes, binaries or other lists, got invalid entry: %Phoenix.HTML.Form{...}
```

That's because `i` will always (as far as I can tell) be list even though it's an single association. So the correct way would be either:

```eex
# Notice that I'm using <%- instead of <%=. If you use <%= here, you will get the same error above.
<%- [i] = inputs_for(f, :child) %>
```

or

```eex
<%= for i <- inputs_for(f, :child) do %>
```

Does that sound right for you or am I missing something?

josevalim:
> Does that sound ok for you or am I missing something? 🤔

Oops, I forgot it always returned lists. So we are good!

josevalim:
> Does that sound ok for you or am I missing something? 🤔

Oops, I forgot it always returned lists. So we are good!

feliperenan:
@josevalim It is ready for review. I have also introduced [this function](https://github.com/phoenixframework/phoenix_html/pull/292/files#diff-149975a43f9e90f073bd077072ba7d50R655) to help with hidden inputs.

lib/phoenix_html/form.ex
```diff
@@ -372,6 +372,56 @@ defmodule Phoenix.HTML.Form do
...
+
+        <%= for friend_form = inputs_for(f, :friends) do %>
+          <%= text_input friend_form, :name %>
+        <% end %>
+      </form>
+
+  A shortcut for `inputs_for(changeset, url, [])`.
+  """
+  @spec inputs_for(t, field) :: list(Phoenix.HTML.Form.t())
+  def inputs_for(form, field), do: inputs_for(form, field, [])
```
josevalim: We can remove this definition and simply use default arguments in the next clause. :)

lib/phoenix_html/form.ex
```diff
@@ -598,6 +648,16 @@ defmodule Phoenix.HTML.Form do
...
   end
 
+  @doc """
+  Generates hidden inputs for the given form.
+  """
+  @spec hidden_inputs_for(t) :: list(Phoenix.HTML.safe())
+  def hidden_inputs_for(form) do
+    for {k, v} <- form.hidden do
+      hidden_input(form, k, value: v)
+    end
```
josevalim: Shall we also use this function [here](https://github.com/phoenixframework/phoenix_html/pull/292/files#diff-149975a43f9e90f073bd077072ba7d50R471)?

lib/phoenix_html/form.ex
feliperenan: We can't  use default arguments in the next clause because `input_for/4`:

```
== Compilation error in file lib/phoenix_html/form.ex ==
** (CompileError) lib/phoenix_html/form.ex:438: def inputs_for/4 defaults conflicts with inputs_for/3
```
I'm not sure what I can do in this case? :thinking:

lib/phoenix_html/form.ex
feliperenan: is that what you mean here?

https://github.com/phoenixframework/phoenix_html/pull/292/files#diff-3a77d7e0ca67e4df7840545f22fa4218R444

lib/phoenix_html/form.ex
```diff
@@ -372,6 +372,56 @@ defmodule Phoenix.HTML.Form do
...
+          # for generating hidden inputs.
+          <%= hidden_inputs_for(company_form) %>
+          <%= text_input friend_form, :name %>
+        <% end %>
+      </form>
+
+  See `inputs_for/4` for the available options.
+  """
+  @spec inputs_for(t, field, Keyword.t()) :: list(Phoenix.HTML.Form.t())
+  def inputs_for(%{impl: impl} = form, field, options) when is_list(options) do
```
josevalim:
```suggestion
  def inputs_for(%{impl: impl} = form, field, options)
      when is_atom(field) and is_list(options) do
```

lib/phoenix_html/form.ex
josevalim: I see. I think you can keep this definition then but you can make the docs very short:

```
@doc """
Same as `inputs_for(form, field, [])`.
"""
```

josevalim:
```suggestion
  def inputs_for(form, field) when is_atom(field), do: inputs_for(form, field, [])
```

lib/phoenix_html/form.ex
josevalim: Sorry, it seems I sent the wrong link. I wanted to propose to use `hidden_inputs_for` here: https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/form.ex#L421-L422

feliperenan: Done!

feliperenan:
lol fixing the wrong rebase! :see_no_evil:

josevalim:

PR #296, opened 2020-03-23
Fix inputs_form spec

lukaszsamson:
v2.14.11 introduced a regression. Now a valid code

```
<%= inputs_for f, :my_field, fn f2 -> %>
    <div class="form-group">
      <%= label f2, :name, class: "control-label" %>
      <%= text_input f2, :name, class: "form-control" %>
      <%= error_tag f2, :name %>
    </div>
<% end %>
```
fails to dialyze with
```
The function call will not succeed.

Phoenix.HTML.Form.inputs_for(_f :: any(), :my_field, (_ -> {:safe, [any(), ...]}))

breaks the contract
(t(), field(), Keyword.t()) :: [t()]

```

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -441,6 +441,7 @@ defmodule Phoenix.HTML.Form do
       within the supplied callback.
 
   """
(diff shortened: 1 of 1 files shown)


josevalim:

PR #300, opened 2020-05-16
Document the index field in Phoenix.HTML.Form

DianaOlympos:
Follow up to the discussion between @chrismccord and @jayjun during the night on the Elixir slack, it seemed this is not meant to be hidden from user, but is not documented.

It now is.


josevalim:

PR #301, opened 2020-05-18
Fix inputs_for/3 is missing is_binary(field) guard

50kudos:
According to 
```elixir
@type field :: atom | String.t()
```
the `inputs_for/3` is missing `is_binary(field)` guard. 

This PR adds the guard so it's consistent with the rest of module's functions which typically have `is_atom(field) or is_binary(field)` guard that allow passing binary field.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -378,7 +378,8 @@ defmodule Phoenix.HTML.Form do
   Same as `inputs_for(form, field, [])`.
   """
   @spec inputs_for(t, field) :: list(Phoenix.HTML.Form.t())
-  def inputs_for(form, field) when is_atom(field), do: inputs_for(form, field, [])
+  def inputs_for(form, field) when is_atom(field) or is_binary(field),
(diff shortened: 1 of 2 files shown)


josevalim:
Please do not include unicode characters in atoms as that will fail in earlier Erlang/OTP versions. :)

50kudos:
Note that the following html also work:

```html
<label for="F✓o]o%b+a'R">Enter your name: </label>
<input type="text" name="F✓o]o%b+a&#39;R" id="F✓o]o%b+a&#39;R" required>
```

When clicking the label, the input gets focused, that mean `F✓o]o%b+a'R` (for) can associate with `F✓o]o%b+a&#39;R` (id).

When submitting a form with the input name, the query string also read correctly.
Screen Shot 2020-05-19 at 1 27 15 PM

josevalim:
The only issue is that we cannot have Unicode in atoms in earlier Erlang versions, and that broke the build. We should probably remove the Unicode characters for now.

josevalim:

PR #304, opened 2020-05-23
Allow extra html attributes with select option generated with prompt

### LINKED ISSUE #303: Allow to pass html attributes to a select option when the option has been generated with prompt
I was trying to reproduce the following html with an extra "disabled" html attribute to the first option that proposes to select a role:
```html
<select id="user_role" name="user[role]">
    <option value="" disabled>Choose your role</option>
    <option value="admin">Admin</option>
    <option value="user">User</option>
</select>
```
The doc proposes a [prompt option](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#select/4-options), so I was trying to use it:
```elixir
select(form, :role, ["Admin": "admin", "User": "user"], prompt: "Choose your role")
```
This options does not allow to generate extra html attribute so the result is the following:
```html
<select id="user_role" name="user[role]">
    <option value="">Choose your role</option>
    <option value="admin">Admin</option>
    <option value="user">User</option>
</select>
```

I find to workarounds:

1. Add the option manually:
```elixir
select(form, :role, [
[value: "", key: "Choose your role", disabled: true],
[value: "admin", key: "Admin"],
[value: "user", key: "User"])
```
2. Use an optgroup to simulate the behavior but it does not select the group label:
```elixir
select(form, :role, ["Choose your role": ["Admin": "admin", "User": "user"]])
```
To be honest I don't like these solutions. I implemented a solution in phoenix_html that allow to pass a list to the `prompt` option but I would like to have your opinion about the usefulness of this implementation before creating a pull request.

So you will be able to use the prompt option as following:
```elixir
select(form, :role, ["Admin": "admin", "User": "user"], prompt: [key: "Choose your role", disabled: true])
```
and it will generate the following code:
```html
<select id="user_role" name="user[role]">
    <option value="" disabled>Choose your role</option>
    <option value="admin">Admin</option>
    <option value="user">User</option>
</select>
```
I use the "key" key to be consistent with the current implementation.

Do you see any drawback?

Malian:
Closes #303 

This PR allows to add extra html attributes to a select option that has been generated with the `prompt` option.

The following code
```elixir
select(:user, :role, ["Admin": "admin", "User": "user"], prompt: [key: "Choose your role", disabled: true])
```
will generate this HTML output:
```html
<select id="user_role" name="user[role]">
    <option value="" disabled>Choose your role</option>
    <option value="admin">Admin</option>
    <option value="user">User</option>
</select>
```

Feel free to ask me for modifications if you think I could better code this behaviour, as this is my first PR in the Elixir world :partying_face:

It also seems that `mix format` corrected a typo in a test file. I can do a separate PR if needed.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1126,12 +1126,16 @@ defmodule Phoenix.HTML.Form do
           <option value="user">User</option>
           </select>
 
+  You can also pass a prompt:
+
       select(form, :role, ["Admin": "admin", "User": "user"], prompt: "Choose your role")
       #=> <select id="user_role" name="user[role]">
           <option value="">Choose your role</option>
           <option value="admin">Admin</option>
           <option value="user">User</option>
           </select>
+
+  And customize the prompt as any other entry:
 
       select(form, :role, ["Admin": "admin", "User": "user"], prompt: [key: "Choose your role", disabled: true])
       #=> <select id="user_role" name="user[role]">
@@ -1164,8 +1168,8 @@ defmodule Phoenix.HTML.Form do
 
   ## Options
 
-    * `:prompt` - an option to include at the top of the options with
-      the given prompt text
+    * `:prompt` - an option to include at the top of the options. It may be
+      a string or a keyword list of atttributes and the `:key`
 
     * `:selected` - the default value to use when none was sent as parameter
 
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1133,6 +1133,13 @@ defmodule Phoenix.HTML.Form do
           <option value="user">User</option>
           </select>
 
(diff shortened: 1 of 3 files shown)


lib/phoenix_html/form.ex
```diff
@@ -1185,6 +1196,24 @@ defmodule Phoenix.HTML.Form do
...
   end
 
+  defp prompt_option(prompt) when is_list(prompt) do
+    {prompt_key, prompt_opts} = Keyword.pop(prompt, :key)
+
+    prompt_key ||
+      raise ArgumentError,
+            "expected :key key when building a prompt select option with a keyword list: #{
+              inspect(prompt)
+            }"
```
josevalim, line 1206:
```suggestion
            "expected :key key when building a prompt select option with a keyword list: " <>
              inspect(prompt)
```

josevalim:

PR #306, opened 2020-06-25
Prefix form id when it is present in form_for/3

feliperenan:
This commit will prefix the form id in its inputs id as form_for/4 does.

It avoid issues when having multiple forms on the same page, you can see more info about it here:

https://github.com/phoenixframework/phoenix_live_view/issues/1014

### DIFF
--- a/lib/phoenix_html/form_data.ex
+++ b/lib/phoenix_html/form_data.ex
@@ -52,11 +52,12 @@ defimpl Phoenix.HTML.FormData, for: [Plug.Conn, Atom] do
   def to_form(conn_or_atom, opts) do
     {name, params, opts} = name_params_and_opts(conn_or_atom, opts)
     {errors, opts} = Keyword.pop(opts, :errors, [])
+    {id, opts} = Keyword.pop(opts, :id)
 
     %Phoenix.HTML.Form{
       source: conn_or_atom,
       impl: __MODULE__,
-      id: name,
+      id: id || name,
       name: name,
       params: params,
       data: %{},
--- a/test/phoenix_html/form_test.exs
+++ b/test/phoenix_html/form_test.exs
@@ -67,7 +67,7 @@ defmodule Phoenix.HTML.FormTest do
     end
   end
 
-  describe "form_for/3 with atomm" do
+  describe "form_for/3 with atom" do
     test "without options" do
       form = form_for(:search, "/", [])
       assert %Phoenix.HTML.Form{} = form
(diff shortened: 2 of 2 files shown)


feliperenan:
We also need to apply this patch in `phoenix_ecto`.

https://github.com/phoenixframework/phoenix_ecto/blob/master/lib/phoenix_ecto/html.ex#L11

test/phoenix_html/form_test.exs
```diff
@@ -261,6 +287,18 @@ defmodule Phoenix.HTML.FormTest do
...
+    test "with id prefix the form id in the input id" do
+      form =
+        safe_to_string(
+          form_for(:search, "/", [params: search_params(), id: "form_id"], fn f ->
+            text_input(f, :key)
+          end)
+        )
+
+      assert form =~
+               ~s(<input id="form_id_search_key" name="search[key]" type="text" value="value">)
```
feliperenan, line 300: As we can see here, the form id is prefixed when using `form_for/4`. Just added this test to ensure that 🙌 

https://github.com/phoenixframework/phoenix_html/blob/700f0e096c06d19c75ccaf4a0ad017bd605dfea6/lib/phoenix_html/form_data.ex#L96

lib/phoenix_html/form_data.ex
```diff
@@ -52,11 +52,14 @@ defimpl Phoenix.HTML.FormData, for: [Plug.Conn, Atom] do
   def to_form(conn_or_atom, opts) do
     {name, params, opts} = name_params_and_opts(conn_or_atom, opts)
     {errors, opts} = Keyword.pop(opts, :errors, [])
+    {id, opts} = Keyword.pop(opts, :id)
+
+    id = if is_nil(id), do: name, else: to_string(id) <> "_#{name}"
```
josevalim, line 57: My suggestion is to use `id || name` here too. :)

feliperenan: Ow, that makes sense. Somehow I got confused thinking the name has been prefixed along with id 😅

josevalim:

lib/phoenix_html/form_data.ex
```diff
@@ -52,11 +52,12 @@ defimpl Phoenix.HTML.FormData, for: [Plug.Conn, Atom] do
   def to_form(conn_or_atom, opts) do
     {name, params, opts} = name_params_and_opts(conn_or_atom, opts)
     {errors, opts} = Keyword.pop(opts, :errors, [])
+    {id, opts} = Keyword.pop(opts, :id)
```
feliperenan, line 55: Hey @josevalim I mess it up. We can't use `Keyword.pop/2` here since we still need to pass opts with id forward 🙈 

The correct would be:

```Elixir
    id = Keyword.get(opts, :id) || name
```

I can open another PR unless you want to fix it right way 😅

feliperenan: PR is open: https://github.com/phoenixframework/phoenix_html/pull/307

feliperenan: https://github.com/phoenixframework/phoenix_ecto/pull/129

PR #309, opened 2020-07-10
Allow %URI{} to be passed to link/2 and button/2 as :to

### LINKED ISSUE #308: `link(name, to: %URI{})` as a valid option
I'll submit a PR if this seems like a good idea, but it would be nice for the `to:` option to accept a `%URI{}` struct as a valid option. My thought is that it would be as simple as doing a `to_string` on the URI.

We could just pass through the string to the validate step, or it could have a slightly optimized version of the validation https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/link.ex#L254

davydog187:
Closes #308

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -239,6 +239,10 @@ defmodule Phoenix.HTML.Link do
     {value, opts}
   end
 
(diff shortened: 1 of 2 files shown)


davydog187:
I should probably update the documentation

josevalim:

PR #315, opened 2020-09-03
Honor the `form` attribute when creating hidden checkbox input

nshafer:
HTML5 allows an `<input>` to specify the form that it is associated with by using the [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefform) attribute. This works fine for every Phoenix.HTML.Form function I've used except `checkbox/3`'s hidden input for the unchecked value. This change simply copies the `form` attribute from the checkbox when it creates the hidden unchecked input, so that it is sent to the correct form.

An example. If I have this template:

```elixir
<%= f = form_for :foo, "/foo", id: "myform" %>
</form>
...
<%= checkbox f, :is_admin, form: "myform" %>
```

it will produce this HTML:

```html
<form action="/foo" id="myform" method="post">
</form>

<input name="foo[bar]" type="hidden" value="false">
<input form="myform" id="foo_bar" name="foo[bar]" type="checkbox" value="true">
```

Which results in the hidden field not being submitted to the proper `<form>` since it is missing the `form` attribute. This fix simply copies it, if it's given, so that the hidden input is posted to the same form.

```html
<input form="myform" name="foo[bar]" type="hidden" value="false">
<input form="myform" id="foo_bar" name="foo[bar]" type="checkbox" value="true">
```

(If you're wondering why you would do this instead of putting the checkbox inside of the `<form></form>` there are several valid reason to not do this, especially since the form is a structural and semantic element in the DOM. In my particular case, I need to have a form for each row of a table, but you can't mix up a `<form>` tag between `<tr>` and `<td>`. So instead I stick the form in the first `<td></td>` then point the inputs in each following cell to that form. I'm doing this with LV even, and with this fix, it works perfectly.)

Thanks!
Nate

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1079,7 +1079,7 @@ defmodule Phoenix.HTML.Form do
       hidden_opts = [type: "hidden", value: unchecked_value]
 
       html_escape([
-        tag(:input, hidden_opts ++ Keyword.take(opts, [:name, :disabled])),
+        tag(:input, hidden_opts ++ Keyword.take(opts, [:name, :disabled, :form])),
         tag(:input, [value: checked_value] ++ opts)
       ])
     else


josevalim:

PR #316, opened 2020-09-18
Use to_iso8601 as the standard implementation for safe dates and times

jeregrine:
Most browsers are fine with with the non-iso-standard to_string we provide in Elixir that does not include the T but some browsers, Safari in particular, will choke and blow up. Using the ISO standard as a default and asking the user to format or choose their own representation is a more clear choice. 

This change does regress and will not use a users configured timezone abbr by default but they can simply call the function to fix that. 

(Please ignore the branch name)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+# v3 (2020-09-18)
+
+* Backwards Incompatible 
+  * By default dates and times will format to the `to_iso8601` functions provided by their implementation.
+
+ 
 ## v2.14.2 (2020-04-30)
 
 * Deprecations
--- a/lib/phoenix_html/safe.ex
+++ b/lib/phoenix_html/safe.ex
@@ -37,7 +37,11 @@ defimpl Phoenix.HTML.Safe, for: NaiveDateTime do
 end
 
 defimpl Phoenix.HTML.Safe, for: DateTime do
-  defdelegate to_iodata(data), to: DateTime, as: :to_iso8601
+  def to_iodata(data) do
+    # Call escape in case someone can inject reserved
+    # characters in the timezone or its abbreviation
+    Plug.HTML.html_escape_to_iodata(DateTime.to_iso8601(data))
+  end
 end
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 # Changelog
 
-## v2.14.2 (2020-04-30)
```
josevalim, line 3: Please do not remove chanelog entries :)

lib/phoenix_html/safe.ex
```diff
@@ -25,23 +25,19 @@ defimpl Phoenix.HTML.Safe, for: BitString do
...
+  defdelegate to_iodata(data), to: NaiveDateTime, as: :to_iso8601
 end
 
 defimpl Phoenix.HTML.Safe, for: DateTime do
-  def to_iodata(data) do
-    # Call escape in case someone can inject reserved
-    # characters in the timezone or its abbreviation
-    Plug.HTML.html_escape_to_iodata(DateTime.to_string(data))
-  end
+  defdelegate to_iodata(data), to: DateTime, as: :to_iso8601
```
josevalim, line 40: I am starting to think this is not desired for DateTime as iso8601 has loss of information. Although it is inconsistent with the other ones... so I am not sure how to proceed.

CHANGELOG.md
jeregrine: doh

lib/phoenix_html/safe.ex
jeregrine: I had similar feelings, I personally thing being consistent and *predictable* are more important in this case. 

The new user won't be setting custom timezone abbreviations, but those that do will know where/how too look to fix this.

josevalim, line 40:
```suggestion
  def to_iodata(data) do
    # Call escape in case someone can inject reserved
    # characters in the timezone or its abbreviation
    Plug.HTML.html_escape_to_iodata(DateTime.to_iso8601(data))
  end
```

josevalim:

jeregrine:
WOOT

Issue #317, opened 2020-09-24
Button data-to query string dropped


mattludwigs:
I was trying to use the `button` function along with path helpers to add a query parameter to the button get request. I noticed it seemed the query parameter was dropped in the logs. After inspecting the element I saw the correct `data-to` string was generated but anytime I pressed the button the backend would not receive the query param.

I might be using the button functionality incorrectly but I haven't found anything yet that would indicate that - but I am very much open to the possibility this is a user error and there is a better way.

I created a little example app that exercises the functionality that I am seeing and how I am using the button function.

https://github.com/mattludwigs/query_string_test

In the [controller](https://github.com/mattludwigs/query_string_test/blob/master/lib/query_string_web/controllers/page_controller.ex) the first pattern match is never hit when using the buttons in the [template](https://github.com/mattludwigs/query_string_test/blob/master/lib/query_string_web/templates/page/index.html.eex).

Dep Versions:

```
phoenix_html v2.14.2
phoenix v1.5.5
```

josevalim:
We build a form behind the scenes and because of that [the query parameters are stripped](https://stackoverflow.com/questions/1116019/submitting-a-get-form-with-query-string-params-and-hidden-params-disappear). Maybe we can automatically convert the query parameters into form inputs but that would require some extra JavaScript code. The other option is to perhaps document this behaviour so maybe it will help avoid further confusion in the future.

mattludwigs:
I figured I was probably using the function incorrectly and there is a better/different way. I just wanted to report this just to make sure it was expected behavior and verify my assumption. It would seem that documentation would be enough. If that sounds okay to you, I should have some time this weekend to add some documentation.

josevalim:
A PR to improve the docs sounds great then!

PR #318, opened 2020-09-27
Add documentation about query parameters for button/2

mattludwigs:
According to the W3C specification the query parameters in the action
field of a form will be stripped when submitting a form with the method
get.

Closes: #317

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -188,6 +188,11 @@ defmodule Phoenix.HTML.Link do
 
   All other options are forwarded to the underlying button input.
 
+  When the `:method` is set to `:get` and the `:to` URL contains query
+  parameters the generated form element will strip the parameters in accordance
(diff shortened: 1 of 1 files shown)


lib/phoenix_html/link.ex
```diff
@@ -188,6 +188,11 @@ defmodule Phoenix.HTML.Link do
 
   All other options are forwarded to the underlying button input.
 
+  When the `:method` is set to `:get` and the `:to` URL contains query
```
josevalim, line 191: I believe this happens regardless of the `:method`, no? Or is it only for `:get` indeed?

mattludwigs: I added some tests for post and delete [here](https://github.com/mattludwigs/query_string_test/commit/8b94f26a395b6736409639b960ec43fd71de4b41) and I am getting the query params from those, so it appears that the params are only striped for `:get`, unless I am miss understanding or my test is making a wrong assumptions.

josevalim:

PR #320, opened 2020-10-07
Respect modifier key clicks for links with :method opt

### LINKED ISSUE #319: Links defined with :method opt do not respect cmd/ctrl clicks
A frequently used browser feature is to click on a link while holding cmd, which opens the link in a new tab while keeping the current tab focused (supported in all major browsers).

When a `:method` opt is added to the `link` helper function, this behavior breaks, and the link opens in the current tab as if cmd wasn't held down. If `target: "_blank"` is added to the `link`, cmd + click behaves as expected regardless of the `:method`. Can be observed in standard templates as well as with LiveView.

Reproducible with:

```ex
<%= link "Link with method", to: Routes.page_path(@conn, :post_link), method: :post %>
```

Happy to look into fixing this if relevant!

jtormey:
Makes it so that the behavior of modifier key clicks (cmd/ctrl + click, shift + click) on links with a `:method` opt more closely matches that of normal links. Does so by adding `target="_blank"` to the hidden form when a modifier key is used.

Preference is given to the existing link target, if present.

## Testing

Chrome: Matches normal links exactly
Safari: cmd + click matches exactly, shift + click opens a new window (instead of adding to reading list)
Firefox: Both cmd + click and shift + click open a focused new tab

The behavior doesn't match normal links exactly, however it tends not to be far off. Whereas before, the link would always result in navigation in the current browser tab.

Closes #319

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -24,7 +24,7 @@
     return input;
   }
 
-  function handleClick(element) {
+  function handleClick(element, targetModifierKey) {
     var to = element.getAttribute("data-to"),
         method = buildHiddenInput("_method", element.getAttribute("data-method")),
         csrf = buildHiddenInput("_csrf_token", element.getAttribute("data-csrf")),
@@ -36,6 +36,7 @@
     form.style.display = "hidden";
 
     if (target) form.target = target;
+    else if (targetModifierKey) form.target = "_blank";
 
     form.appendChild(csrf);
     form.appendChild(method);
@@ -58,7 +59,7 @@
       }
 
       if (element.getAttribute("data-method")) {
-        handleClick(element);
+        handleClick(element, e.metaKey || e.shiftKey);
         e.preventDefault();
         return false;
       } else {


josevalim:
Hi @jtormey, thanks for the PR! A colleague just pointed out that this may not work as intended in many cases. For example, there is no guarantee there is a GET URL matching the POST URL. So given this can actually be broken behaviour to many, I think we should somehow make this opt-in but I am not sure how.

jtormey:
Sure! Happy to contribute.

I'm not sure I understand why there needs to be a matching GET URL. The link leads to a new tab, but the form action is the same as if it were in the same tab (including the HTTP method), thanks to the `target="_blank"`.

So `method: :post` or `method: :delete` will both trigger a POST request in the new tab (or window), along with the proper `_csrf_token` and `_method` data.

jtormey:
As a quick example, here is an example link:

```ex
  <%= link "Link POST", to: Routes.page_path(@conn, :post_link), method: :post %>
```

When clicked (using these changes) the following router scope handles it the same no matter how it's clicked (with or without a modifier key):

```ex
  scope "/", LinkClickWeb do
    pipe_through :browser

    get "/", PageController, :index
    post "/post-link", PageController, :post_link
  end
```

josevalim:
Ah, we misunderstood it then. Thank you so much! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

PR #322, opened 2020-11-20
Correct documentation of Link.button/2

dotdotdotpaul:
It does not (no longer?) generates an actual HTML form, but rather uses
Javascript to build one dynamically and submit.


josevalim:

PR #330, opened 2021-01-21
Add docs and example for Form.label/4

lirossarvet:
Just a small documentation update! I was really confused about how to use something like `gettext` to translate labels. So I wanted to add a little documentation about how to actually specify the text to prevent field inference + an example of it.


josevalim:

Issue #276, opened 2019-10-29
Proposal: conditionally setting css classes


wojtekmach:
It's pretty common to end up with code like this:

```eex
<%= link_to "Home", "/", class: (["button"] ++ (if path == "/", do: ["active"], else: []) %>
```

I'd like to propose to allow passing a map of `class => true|false` to conditionally include a given class instead:

```eex
<%= link_to "Home", "/", class: %{"button" => true, "active" => path == "/"} %>
```

Another idea would be:

```eex
<%= link_to "Home", "/", class: ["button", {"active", path == "/"}] %>
```

wojtekmach:
I think this syntax would be the nicest:

```elixir
class: ["button", "active": path == "/"]
```

but the problem is it creates an atom so if someone does this:

```elixir
class: ["button", "user_#{id}": path == "/"]
```

they could blow up their atom table.

I agree what I proposed is a bit awkward to use, curious if folks come up with other ideas!

tomekowal:
I have a similar but slightly different problem and I was also thinking about potential solutions.
I have lots of tables in one of my projects and wanted to create view helpers for them but they often differ slightly.

By default I want this:

```html
<table class="table">
```

But in one place I need:

```html
<table calss="table table-striped">
```

and in other I need:

```html
<table class="table table-dark">
```

I imagined a solution that would allow for composing HTML before producing io list like this.

```elixir
table() |> add_class("table-dark")
```

or for @wojtekmach example

```elixir
<%= link_to("Home", "/", class: "button") |> maybe_add_active_class(path, "/") %>
```

That is currently impossible because all `Phoenix.HTML` functions return the io lists directly. For composition to work, we would need an intermediate data structure that we can construct step by step and that implements `Phoenix.HTML.Safe`

I started a discussion on Elixir Forum and presented my reasoning
https://elixirforum.com/t/would-you-find-composablehtml-library-useful

Unfortunately, it would require rewriting `Phoenix.HTML.{Link, Form, Tag}` to support composition, so you may find that option not viable.

On the other hand, this would help keep the templating language clean and simple.

wojtekmach:
Looking at it again, I think this is actually not bad:

```elixir
iex> active? = false
iex> link("Home", to: "/", class: ["btn", active? && "active"]) |> Phoenix.HTML.safe_to_string() |> IO.puts()
<a class="btnactive" href="/">Home</a>
```

[...]

PR #331, opened 2021-02-08
Add Tag.attributes_escape/1

msaraiva:
This PR adds a new function called `attributes_escape/1` that will be used by the new template engine I'm working on. I'm not sure this function should be placed in the `Tag` module since neither `tag` nor `content_tag` use it. However, I believe they should use it for consistency. The problem is that, currently, the implementation sorts the attributes, which makes no sense for the new engine as it generates groups of attributes separately based on their types (static key/value, static key + dynamic value and dynamic key/value). 

I can either move this function to someplace else or change the implementation of `tag` and `content_tag` to use it. The latter might break existing code if, for any reason, the order of the attributes was taken into account, for instance, in tests.

### DIFF
--- a/lib/phoenix_html/tag.ex
+++ b/lib/phoenix_html/tag.ex
@@ -8,7 +8,8 @@ defmodule Phoenix.HTML.Tag do
 
   import Phoenix.HTML
 
-  @tag_prefixes [:aria, :data]
+  @special_attributes ["data", "aria", "class"]
+
   @csrf_param "_csrf_token"
   @method_param "_method"
 
@@ -49,7 +50,7 @@ defmodule Phoenix.HTML.Tag do
   def tag(name), do: tag(name, [])
 
   def tag(name, attrs) when is_list(attrs) do
-    {:safe, [?<, to_string(name), build_attrs(attrs), ?>]}
+    {:safe, [?<, to_string(name), build_attrs(attrs) |> Enum.sort() |> tag_attrs(), ?>]}
   end
 
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/tag.ex
```diff
@@ -87,16 +88,46 @@ defmodule Phoenix.HTML.Tag do
...
+  Escapes a list of attributes, returning iodata.
+
+      iex> attributes_escape(title: "the title", id: "the id", selected: true)
+      {:safe,
+       [
+         [32, "title", 61, 34, "the title", 34],
+         [32, "id", 61, 34, "the id", 34],
+         [32, "selected"]
+       ]}
+
```
josevalim, line 106: We should document that, opposite to `tag` and `content_tag`, the attributes are not sorted.

josevalim (approved):
Just one comment and we can ship it!

lib/phoenix_html/tag.ex
```diff
@@ -96,6 +96,9 @@ defmodule Phoenix.HTML.Tag do
   @doc """
   Escapes a list of attributes, returning iodata.
 
+  Pay attention that, unlike `Phoenix.HTML.Tag.tag/2` and `Phoenix.HTML.Tag.content_tag/2`, this
+  function does not sort the attributes.
```
josevalim, line 100:
```suggestion
  Pay attention that, unlike `tag/2` and `content_tag/2`, this
  function does not sort the attributes.
```

josevalim:

msaraiva:
@josevalim regarding accepting `class: [active: path == "/"]` as suggested by @wojtekmach in https://github.com/phoenixframework/phoenix_html/issues/276#issuecomment-547437728, I don't see why we shouldn't accept it. After all, the user can do stuff like `"[user_#{id}": ...]` on basically any function that accepts an atom or a keyword list. I don't see why we should treat this differently. Classes are, by nature, mostly not dynamic and when they are, the set of possible classes are still finite/predefined, not based on some infinite id. I actually never saw anyone using classes based on ids.

PR #334, opened 2021-04-18
Use Github Actions instead of TravisCI

dolfinus:
TravisCI site shows this notice:
"Please be aware travis-ci.org will be shutting down in several weeks"

Let's move CI to Github Actions just like the phoenix repo. This is an example of a successful workflow run: https://github.com/dolfinus/phoenix_html/actions/runs/760751218

And also I've updated README file a bit.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    strategy:
+      matrix:
+        elixir:
+        - "1.7"
+        - "1.8"
+        - "1.9"
+        - "1.10"
+        - "1.11"
+        otp:
```
dolfinus, line 39: There is no OTP 19 because it ubuntu-18.04 and ubuntu-20.04 were updated fex months ago and become incompatible with an installer. Also phoenix repo CI/CD is running tests only for 21-23 OTP versions, so there is actually no need to test depending repo against older OTP versions.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
+name: Check
+
+on:
+  push:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  check_duplicate_runs:
```
dolfinus, line 10: There is an annoying bug of Github Actions - when someone creates a new branch and pushes some commits here, and then creates a pull request, there are 2 copies of workflow running - one for a push event and another for created PR.
This step prevents the second copy of workflow to be run - if tests are already passed for push event then they will be not run twice while PR is created. Also they will not be run if source code hasn't been changed.

josevalim:
Can you please revert the changes to the README as they are unrelated? Thank you. :) I will fix the heading tho directly on master.

dolfinus:
Ok, fixed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        - elixir: "1.8"
+          otp: "23.3"
+        - elixir: "1.9"
+          otp: "23.3"
+        - elixir: "1.10"
+          otp: "20.3"
+        - elixir: "1.11"
+          otp: "20.3"
+        - elixir: "1.11"
+          otp: "21.0"
```
wojtekmach, line 56: I don't think we need such wide build matrix for this particular project. I like to build on just the oldest supported and newest available Elixir/OTP combo and that tends to be good enough to find compatibility bugs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+    - name: Restore deps and _build cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          deps
+          _build
+        key: deps-${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}-git-${{ github.sha }}
+        restore-keys: |
+          deps-${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+          deps-${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}
```
wojtekmach, line 79: is this line necessary? (L78 of course makes sense)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+    - name: Remove compiled application files
+      run: mix clean
+
+    - name: Compile dependencies
+      run: mix compile
+      env:
+        MIX_ENV: test
+
+    - name: Run tests
+      run: mix test
```
wojtekmach, line 93: consider adding some more static analysis checks, we use the following on our nimble_* projects:

https://github.com/dashbitco/mox/blob/036f1aae4015178e4a732aafb39e95362f0ad30e/.github/workflows/ci.yml#L36:L47

.github/workflows/ci.yml
wojtekmach: This is a neat idea to avoid duplicate work. I don't think we need it anymore though, given we have this:

```yaml
on:
   push:
   pull_request:
     branches:
       - master
```

that is, we'd only build if something is a PR or it's a push to master, we'd never trigger two runs for a PR branch.

.github/workflows/ci.yml
dolfinus: Fixed.

.github/workflows/ci.yml
dolfinus: Ok, I've added steps of checking code format and unused dependencies in `mix.lock` file.

wojtekmach:
Thanks @dolfinus!

.github/workflows/ci.yml
dolfinus: `actions/cache` does not upload updated cache files if cache was downloaded using primary key (without fallback to restore-keys).
But sometimes it can be very useful. For example, if someone changes workflow file or adds tools like dialyzer which can update `deps` or `_build` folders content without changing `mix.lock` hash.

Adding `-git-${{ github.sha }}` to the key forces the action to upload `deps` and `_build` folders to the cache on every commit.
This does not increase cache search time or something like that.

.github/workflows/ci.yml
dolfinus: Well, if you don't use internal PRs from feature branch to a master, then this step is not needed.

.github/workflows/ci.yml
wojtekmach: > Adding -git-${{ github.sha }} to the key forces the action to upload deps and _build folders to the cache on every commit. This does not increase cache search time or something like that.

Agreed there's no downside when fetching the cache but I think it would be good to not do unnecessary cache writes either. I'd personally hold off on this change until we run into a particular issue that it would have solved, but not strong feelings.

TIL we can actually remove L78 as L79 covers it, the pattern is greedy:

> `npm-foobar-d5ea0750` matches a specific hash.
> `npm-foobar-` matches cache keys prefixed with npm-foobar-.
> `npm-` matches any keys prefixed with npm-.

(https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)

dolfinus: Ok, removed.

josevalim:

PR #335, opened 2021-04-18
Properly case EEx

eksperimental:


lib/phoenix_html/engine.ex
```diff
@@ -184,7 +184,7 @@ defmodule Phoenix.HTML.Engine do
 
       :error ->
         raise ArgumentError, """
-        assign @#{key} not available in eex template.
+        assign @#{key} not available in EEx template.
```
josevalim, line 187:
```suggestion
        assign @#{key} not available in template.
```

josevalim:

PR #336, opened 2021-04-18
Run mix format

eksperimental:


eksperimental:
Wondering if we should add a `mix format --check-formatted` to the CI

josevalim:
Fixed in master by an earlier PR, thanks!

PR #338, opened 2021-04-19
Fix CI

dolfinus:
There was an error with CI workflow which I haven't faced with in my fork because of cached dependencies, but it raised after merging into original repo:
https://github.com/phoenixframework/phoenix_html/runs/2378068957

Here is a fix. Also I fixed one failed test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,12 +41,15 @@ jobs:
         restore-keys: |
           deps-${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}
 
+    - name: Install dependencies
+      run: mix deps.get --only test
+
     - name: Check source code format
       run: mix format --check-formatted
       if: ${{ matrix.lint }}
 
     - name: Check unused dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -45,8 +45,11 @@ jobs:
       run: mix format --check-formatted
       if: ${{ matrix.lint }}
 
+    - name: Install dependencies
+      run: mix deps.get
```
mcrumm, line 49:
```suggestion
    - name: Install dependencies
      run: mix deps.get --only test
```

Let's also move this above the format step. @wojtekmach has been updating CI for a lot of Elixir packages and has a nice template we can steal :) here's an example: dashbitco/nimble_parsec#87

mcrumm (commented):
Hi @dolfinus– thanks for the fix, and for the original PR making the switch to Actions. I left a comment for one tiny change. Also if you rebase the other test has been fixed :)

.github/workflows/ci.yml
dolfinus: Fixed.

mcrumm:
Thank you! <3

PR #339, opened 2021-05-13
Allow DateTime to be used with datetime_local_input/3

davydog187:
Paired with @supersimple on this one. 

Found this via testing with the Surface [DateTimeLocalInput](https://surface-ui.org/builtincomponents/DateTimeLocalInput) component. 

As for the expected behavior, this will not take any timezone into account. This seems reasonable to me, but happy to change based on feedback.


josevalim:
I believe the current behavior is intentional. Ignoring the timeZone should be a explicit decision, I recommend converting to naive explicitly before. :)

davydog187:
@josevalim the problem is that if you give it a datetime, it silently breaks. Do you recommend raising an error then?

davydog187:
Net effect of passing a `DateTime` right now is that it will call `to_string`, which will end up as an improper `value` of the `input` field.

davydog187:
@josevalim did you mean to merge?

josevalim:
Yes, I forgot to write it down though. I figured out that, if the input is a DateTime, it must have been converted to the user local timezone, which may or may not be stored separately, so I think it should be fine to show the DateTime as is rather than asking the users to convert to native.

davydog187:
I thought the same, thank you!

PR #344, opened 2021-08-06
Fix label do block example

### LINKED ISSUE #343: label do block example is incorrect
As we can see in the docs, the label do block example both show the same result. 
I assume this is a simple copy/paste error.

<details open>
<summary>Docs for label do block</summary>
Screen Shot 2021-08-06 at 4 31 36 PM

</details>

tomkonidas:
Closes #343 

This PR updates the example to match the real returned value of the label do block.


mcrumm:
Thanks!

PR #352, opened 2021-09-09
Feature: disable links until JS properly loaded

benoawfu:
This PR disables all links with a method other than `:get` until PhoenixHTML is loaded (= the expected behaviour is ready), to avoid issueing a wrong GET call if somehow phoenixJS is not yet loaded (latency, JS/bundling error, anything that could prevent phoenix_html.js to load).

This way we ensure nobody can hit the wrong GET route when we intended to use only the POST/PUT/DELETE one. For instance a POST might redirect to the matching GET route, in that case you might never know that your POST didn't go through and you just clicked on a link to fetch only the GET route.

I've handled the case where you wanted to create a link disabled, to ensure we don't enable it automatically in JS.

Updated documentation, unit tests and JS part tested in real life.

### DIFF
--- a/lib/phoenix_html/link.ex
+++ b/lib/phoenix_html/link.ex
@@ -27,7 +27,7 @@ defmodule Phoenix.HTML.Link do
 
       # If you supply a method other than `:get`:
       link("delete", to: "/everything", method: :delete)
-      #=> <a href="/everything" data-csrf="csrf_token" data-method="delete" data-to="/everything">delete</a>
+      #=> <a href="/everything" data-csrf="csrf_token" data-method="delete" data-to="/everything" data-load-disabled="true">delete</a>
 
       # You can use a `do ... end` block too:
       link to: "/hello" do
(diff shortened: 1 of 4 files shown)


Gazler:
Hi, thanks for the PR. Functionally, I think this could work in a specific application (running the JS you included in `app.js`), however I have concerns merging this as it could cause an issue with LiveView. For example:

If you have `Routes.link(to: some_route(socket, @dynamic_id), method: :post)` in your live_view, then if `@dynamic_id` changes, the link would be permanently disabled, as the re-rendered HTML would clobber the removal of the `disabled` attribute triggered from the element being replaced on LiveView. If we were to merge this, we'd have to consider this in LiveView too.

benoawfu:
I see, yes I didn't consider LiveView using it, so it would need some rework there for sure.

If anyone sees some value in this feature -as I do-, I'll be more than happy if anyone is willing to help to get there!

josevalim:
I will close this one for now, as it has gone stale. :)

PR #360, opened 2021-11-01
support rendering hidden inputs for list of values

### LINKED ISSUE #359: Rendering array field as hidden inputs
I'm missing a way to automatically render an array field as hidden inputs. What I'd like to be able to do is this:

```elixir
<%= form_for :some_form, "/", [hidden: [some_field: ["one", "two", "three"]]], fn f -> %>
  <%= hidden_inputs_for f %>
<% end %>
```

Which should render to:

```html
<form action="/" method="post">
  <input type="hidden" id="something_0" name="something[]" value="one" />
  <input type="hidden" id="something_1" name="something[]" value="two" />
  <input type="hidden" id="something_2" name="something[]" value="three" />
</form>
```

Would it make sense to modify `hidden_inputs_for/1` to behave that way? If so, I can open a PR.

Background: I'm currently working on tighter integration between [flop](https://github.com/woylie/flop), [flop_phoenix](https://github.com/woylie/flop_phoenix) and `Phoenix.HTML.Form` by implementing `Phoenix.HTML.FormData` for a struct. The goal is to make setting up filter forms based on the library structs easier. To work well with the sortable table and pagination components, parameters like the page size etc. are added to the `:hidden` field in `form_for/2`, so that all of those parameters can be rendered with a single call to `hidden_inputs_for/1`. For sorting however, an `:order_by` field is used, which is a list. I tried modifying the protocol implementation to include that list as hidden fields as well, but I wasn't successful, which means at the moment I would have to add a modified `hidden_inputs_for/1` function to the my library. I'd rather have this part of Phoenix.HTML.

woylie:
Something like this?

resolves #359

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -631,9 +631,26 @@ defmodule Phoenix.HTML.Form do
   """
   @spec hidden_inputs_for(t) :: list(Phoenix.HTML.safe())
   def hidden_inputs_for(form) do
-    for {k, v} <- form.hidden do
-      hidden_input(form, k, value: v)
+    Enum.flat_map(form.hidden, fn {k, v} ->
+      hidden_inputs_for(form, k, v)
+    end)
+  end
+
+  defp hidden_inputs_for(form, k, values) when is_list(values) do
+    id = input_id(form, k)
+    name = input_name(form, k)
+
+    for {v, index} <- Enum.with_index(values) do
+      hidden_input(form, k,
+        id: id <> "_" <> Integer.to_string(index),
+        name: name <> "[]",
+        value: v
+      )
     end
+  end
+
+  defp hidden_inputs_for(form, k, v) do
+    [hidden_input(form, k, value: v)]
   end
 
   @doc """
--- a/test/phoenix_html/form_test.exs
+++ b/test/phoenix_html/form_test.exs
@@ -509,6 +509,17 @@ defmodule Phoenix.HTML.FormTest do
 
       assert hidden_inputs_for(form) == [hidden_input(form, :id, value: 1)]
     end
+
+    test "generates hidden fields for lists from the given form" do
+      form = %{form_for(conn(), "/") | hidden: [field: ["a", "b", "c"]]}
+
+      assert hidden_inputs_for(form) ==
+               [
+                 hidden_input(form, :field, name: "field[]", id: "field_0", value: "a"),
+                 hidden_input(form, :field, name: "field[]", id: "field_1", value: "b"),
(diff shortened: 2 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -631,11 +631,28 @@ defmodule Phoenix.HTML.Form do
...
+      hidden_inputs_for(form, k, v)
+    end)
+  end
+
+  defp hidden_inputs_for(form, k, values) when is_list(values) do
+    id = input_id(form, k)
+    name = input_name(form, k)
+
+    for {v, index} <- Enum.with_index(values) do
+      hidden_input(form, k,
```
josevalim, line 644: I am thinking it may be better to pass `id: false` on all of them?

woylie: Wouldn't that be the only inputs that would be rendered without an `id` then?

josevalim: Yeah, but the odds of customizing them is zero? And this reduce any potential conflict when passing a value as a list. Otherwise the combination of `#{field}_#{value}` may conflict with some `#{field}`?

woylie: The ID would be `#{field}_#{index}`, but you're right, you probably won't use the ID anywhere.

woylie: I changed it.

lib/phoenix_html/form.ex
```diff
@@ -631,11 +631,27 @@ defmodule Phoenix.HTML.Form do
...
+        id: false,
+        name: name <> "[]",
+        value: v
+      )
     end
   end
 
+  defp hidden_inputs_for(form, k, v) do
+    [hidden_input(form, k, value: v)]
```
josevalim, line 652:
```suggestion
    [hidden_input(form, k, id: false, value: v)]
```

josevalim: Hrm, now I see this may be a breaking change. I guess people could be using this to read a DOM value or something? 🤔 Should we keep the `id: false` only for lists then?

woylie: Yes, this might break things. Some people like using IDs in tests as well.

lib/phoenix_html/form.ex
woylie: Although I must say that I love consistency, so this hurts a bit.

josevalim: Yeah, agreed. Sorry about it, can we please revert?

woylie: Sure 😄

josevalim:

PR #362, opened 2021-11-14
Support nested phx-* attributes

hubertlepicki:
Similar to existing aria-* and data-* attributes support, allow for
passing nested phx-* attributes to tags.


josevalim:

PR #366, opened 2021-12-23
Have `checkbox` use `value` only when `checked` needs to be determined

marcandre:
This is particularly useful for fields of type `{:array, Ecto.Enum}` which would fail to be converted to HTML:

```
** (ArgumentError) lists in Phoenix.HTML and templates may only contain integers representing bytes, binaries or other lists, got invalid entry: :a
```

I imagine that converting the items of `value` to HTML and setting `checked` according to if `checked_value in value` would be deemed too much.

I'm happy to compute `checked` myself, but I believe that having to pass a bogus `value` parameter for the code not to crash is a bug.

Note that this PR introduces a potential behavior change for:
```
checkbox(..., value: "a", checked_value: "a", checked: false)
````

Current behavior overrides the `checked` to `true` while this PR maintains the explicitly given value of `false`. It seems better to me, but not by much as this example seems illogical.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -1055,16 +1055,14 @@ defmodule Phoenix.HTML.Form do
     # We html escape all values to be sure we are comparing
     # apples to apples. After all we may have true in the data
     # but "true" in the params and both need to match.
-    value = html_escape(value)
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #367, opened 2021-12-27
Incorrect documentation in a forms example(?)


nyaray:
I don't know if I'm doing something wrong or if I found an error in the docs of phoenix_html 3.2.0. I'm having issues with a form because of CSRF stuff. Well, I'm trying to do what's in the docs (https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#module-with-changeset-data) to let the framework handle CSRF and am getting an unexpected error (well, actually a warning).

The warning I'm getting is

```
MyApp.Router.Helpers.user_path/2 is undefined or private. Did you mean:

      * page_path/2
      * page_path/3
```

I've been able to fix the issue by updating my template to call `page_path/2` instead.

It's a bit jarring to only get a compile warning (who reads those, anyway, amirite?) and then have the page 500 when accessing the page that presents the form, although I understand why it happens, hehe. I've set warnings as errors in my compile options now, which I'd forgotten to do 😅.

josevalim:
Thank you! Closing in favor of the PR!

PR #368, opened 2021-12-27
Fix incorrect examples in docs.

nyaray:
This addresses issue #367

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -41,7 +41,7 @@ defmodule Phoenix.HTML.Form do
   Now a `@changeset` assign is available in views which we
   can pass to the form:
 
-      <%= form_for @changeset, Routes.user_path(@conn, :create), fn f -> %>
+      <%= form_for @changeset, Routes.page_path(@conn, :create), fn f -> %>
         <label>
           Name: <%= text_input f, :name %>
         </label>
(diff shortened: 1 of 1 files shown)


josevalim:
Your application is not guaranteed to have `page_path` either. I mean, a new application does have a `page_path` but only the `:index` action. The main point is that you should those by your own routes... perhaps we should leave the example as is but add a paragraph saying you should replace `user_path` by the relevant resource/route?

nyaray:
OK, at first I was _very_ confused, because I didn't realise that `user_` in the function name came from the entity one was dealing with. The prefix in the example comes from the fact that it's related to the `UserController`, right?

What I don't understand is how I arrive at the conclusion that the first part of the controller's name is the prefix of the function I want to call in my template. I think that is what needs to be clarified in the extra paragraph you're talking about, which is a good idea, since my edits were made under an incorrect assumption about how Router helper functions were generated.

josevalim:
@nyaray exactly. For forms there is usually something like

    resources :users, UserController

Which means there is a `user_path/2` helper with several actions, `:create` being one of them.

nyaray:
Is that `resources :users ...` generated by a `use` declaration somewhere? I don't have it anywhere, according to my search-entire-project function. If so, that should probably also be part of the explanation.

josevalim:
If you use `mix phx.gen.html`, then it asks you to add it to the router. But in a nutshell, it can be any route previously defined in your router.

josevalim:
Closing for now. If you can think of other improvements, we will be glad to reopen. Thanks!

PR #369, opened 2021-12-31
clarify link do-end doc

restlessronin:


restlessronin:
The example showing how link works with a do-end block was missing the output. When I was first looking at the documentation, i missed the utitilty of the block. I think adding the output makes it much clearer.

josevalim:

restlessronin:
🙏🏾
❤️ your work

PR #375, opened 2022-02-18
Use artificial button click instead of form.submit

### LINKED ISSUE #374: Forms are not really submitted
We use hotwire/turbo for one of our projects and turbo relies on the convention that a form triggers a submit.

In phoenix html however the form submit does not really trigger a submit event because `form.submit()`[1] somwhat unintuitively does _not_ raise a submit event. To do that we'd need to use `form.requestSubmit()`[2].

The line we'd need to adjust is this one: https://github.com/phoenixframework/phoenix_html/blob/3214719a297b89788dc18f4adf573f7483513e07/priv/static/phoenix_html.js#L44

Would you be open to this change? In that case I'd make a pull request.

Thanks! :-)

[1]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit
[2]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit

mradke:
Fixes #374

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -29,6 +29,7 @@
         method = buildHiddenInput("_method", element.getAttribute("data-method")),
         csrf = buildHiddenInput("_csrf_token", element.getAttribute("data-csrf")),
         form = document.createElement("form"),
+        submit = document.createElement("input"),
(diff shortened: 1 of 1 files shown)


josevalim:

Issue #372, opened 2022-01-16
Phoenix.HTML goes slim


josevalim:
With HEEx, many of the functions in this module have lost their purpose. Therefore, our goal is to remove redundant functionality from `Phoenix.HTML`. In particular, `Phoenix.HTML.Format`, `Phoenix.HTML.Link`, and `Phoenix.HTML.Tag` should be deprecated. To do so, we need to move some of the required functionality around:

  * [x] `Phoenix.HTML.Tag.attributes_escape` will be moved to `Phoenix.HTML`
  * [x] `Phoenix.HTML.Link.button` will be broken into `Phoenix.HTML.link_attributes/2`
  * [x] Import `Phoenix.Controller.get_csrf_token()` to `use MyAppWeb, :view` and use it instead of `csrf_meta_tag` in templates
  * [x] `Phoenix.HTML.Link.link/2` should be moved to `Phoenix.LiveView.Helpers.link/1` function component with a unified API
  * [x] Change `mix phx.new` to only import `Phoenix.HTML` and `Phoenix.HTML.Form`
  * [x] Soft deprecate `Phoenix.HTML.{Format, Link, Tag}`

There are some open discussion points:

  * [ ] Should we have a replacement for `Phoenix.HTML.Form`?
  * [ ] If we do have one, should `Phoenix.HTML` even exist? Perhaps it should be a `phoenix_html_safe` project and the existing `Phoenix.HTML` module is part of `Phoenix.View.HTMLHelpers` or similar?

leandrocp:
Hey @josevalim is this issue open for PRs? I could give a hand.

josevalim:
I think we still have to figure out some of those but we can definitely go ahead with "Move Phoenix.HTML.Tag.csrf_token_value/0-1 to Phoenix.HTML (convert csrf_meta_tag and csrf_input_tag to examples)" and then change the generators to use csrf_token_value.

leandrocp:
> I think we still have to figure out some of those but we can definitely go ahead with "Move Phoenix.HTML.Tag.csrf_token_value/0-1 to Phoenix.HTML (convert csrf_meta_tag and csrf_input_tag to examples)" and then change the generators to use csrf_token_value.

Seems like "Move Phoenix.HTML.Tag.tag/2 and content_tag/3 to Phoenix.HTML" has no blocks and should be good for a PR too.

josevalim:
Most of it is done!

sevab:
[...]

PR #378, opened 2022-04-23
Move Phoenix.HTML.Tag.tag/content_tag to Phoenix.HTML

leandrocp:
Ref https://github.com/phoenixframework/phoenix_html/issues/372


josevalim:
I am not sure we want to do this one yet. So we have to wait. :)

josevalim:
To be clear, only the csrf_token_value one is 100% sure at this point!

leandrocp:
> To be clear, only the csrf_token_value one is 100% sure at this point!

No worries! Closing it for now

PR #381, opened 2022-05-11
Fix `checkbox` generating wrong `id` parameter for multi checkbox case

marcandre:
Previously, the same ID would be generated for all checkboxes.


josevalim:
Given this is the third request in 1 month for this feature, we should probably do it, yeah. :)

josevalim:

PR #382, opened 2022-05-12
Add locals_without_parens (for Phoenix.LiveView.HTMLFormatter)

TheFirstAvenger:
Now that Phoenix.LiveView.HTMLFormatter can format `heex` templates, we should export the functions that can appear in a template that we don't want to force parens on. We can then import `:phoenix_html` in our project `.formatter.exs` files. Sister PR: https://github.com/phoenixframework/phoenix_live_view/pull/2027


josevalim:
We generally list all of those on Phoenix... but also note that we are moving away from the form functions anyway, so we are closing this. :)

PR #384, opened 2022-06-03
Use `defdelegate` for HTML.Safe implementations of Integer and Float

optikfluffel:
While looking at the changes made in #383 I noticed `defdelegate` being used in the implementations of `BitString`, `Time`, `Date` and `NaiveDateTime` and thought it would be nice if the one for `Integer` would do the same.

And, since the current implementation for `Float` is just the code of [`Float.to_string/1`](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/float.ex#L562) copied over, I delegated to that one too.

Feel free to close if you don't like it, just a quick thought I had. ☺️


josevalim:

PR #385, opened 2022-06-14
Allow class list to be nested

LostKobrakai:
Allow for nested class lists.

Context: https://github.com/phoenixframework/phoenix_live_view/issues/2066


josevalim:

PR #388, opened 2022-08-16
Expand note on :errors

nathanl:
Also opened a PR to link here in the Ecto docs: https://github.com/elixir-ecto/ecto/pull/3980

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -62,7 +62,6 @@ defmodule Phoenix.HTML.Form do
   automatically show up as selected in the form.
 
   ### A note on `:errors`
-  
   Even if `changeset.errors` is non-empty, errors will not be displayed in a
   form if [the changeset
   `:action`](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-changeset-actions)


lib/phoenix_html/form.ex
```diff
@@ -62,19 +62,26 @@ defmodule Phoenix.HTML.Form do
   automatically show up as selected in the form.
 
   ### A note on `:errors`
-
-  If no action has been applied to the changeset or action was set to `:ignore`,
-  no errors are shown on the form object even if the changeset has a non-empty
-  `:errors` value.
+
```
josevalim, line 65:
```suggestion

```

nathanl:
On August 16 in the Elixir Slack, @chrismccord said about this PR:
> the doc changes may be a bit too specific because the form errors are shown conditionally in the template by reading the changeset directly in the cast of dead views

I said:
> You mean here? https://github.com/phoenixframework/phoenix_ecto/blob/2cf35cf2e441416bfb82b72aafc1659a12873cd2/lib/phoenix_ecto/html.ex#L292-L294
> I don't get why that makes the doc changes too specific - can you elaborate?

josevalim:

PR #392, opened 2022-10-25
Clarifying note

nathanl:
Can somebody help me write a better or more complete explanation? I used `hidden_input f, :id` inside an `inputs_for` and didn't get that I needed to use `hidden_inputs_for` so that `:id` would not be included on dynamically generated rows for adding new records.


josevalim:
It was made public for the version of inputs_for without anonymous function, but such API should be deprecated, so I would deprecate hidden_inputs_for too. I will push something soon.

PR #400, opened 2023-01-28
Use field atom for input_value

mcrumm:
Fixes implementations (like phoenix_ecto) that expect a field atom to derive value.

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -284,7 +284,7 @@ defmodule Phoenix.HTML.Form do
        form: form,
        id: input_id(form, field_as_string),
        name: input_name(form, field_as_string),
-       value: input_value(form, field_as_string)
+       value: input_value(form, field)
      }}
   end
 
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -384,11 +384,9 @@ defmodule Phoenix.HTML.Form do
         field
       )
       when is_atom(field) do
-    as_string = Atom.to_string(field)
-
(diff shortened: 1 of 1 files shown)


lib/phoenix_html/form.ex
```diff
@@ -284,7 +284,7 @@ defmodule Phoenix.HTML.Form do
        form: form,
        id: input_id(form, field_as_string),
        name: input_name(form, field_as_string),
-       value: input_value(form, field_as_string)
+       value: input_value(form, field)
```
mcrumm, line 287: This change may not be strictly necessary– it raised when using map access (`f[:name]`) with `<.form :let={f} for={@changeset}>` in LiveView.

lib/phoenix_html/form.ex
```diff
@@ -384,11 +384,9 @@ defmodule Phoenix.HTML.Form do
         field
       )
       when is_atom(field) do
-    as_string = Atom.to_string(field)
-
     impl1 != impl2 or id1 != id2 or name1 != name2 or
       Keyword.get_values(errors1, field) != Keyword.get_values(errors2, field) or
-      impl1.input_value(source1, form1, as_string) != impl2.input_value(source2, form2, as_string)
+      impl1.input_value(source1, form1, field) != impl2.input_value(source2, form2, field)
```
mcrumm, line 389: Without this change LiveView raises:

```
(phoenix_live_view 0.18.11) lib/phoenix_live_view/engine.ex:1197: Phoenix.LiveView.Engine.nested_changed_assign?/4
```

josevalim:

PR #404, opened 2023-02-03
Add field to FormField struct

mcrumm:


josevalim:
I think we should merge this but for completeness we will eventually have to support string keys and then field will have to be a string too, but that’s ok.

josevalim:

PR #408, opened 2023-02-15
Allow string keys in form access

sodapopcan:
Hello,

I'm just dipping my toes into contributing outside of just doc fixes so I thought I'd grab [this issue](https://github.com/phoenixframework/phoenix_live_view/issues/2438).  I got the insight from a comment made by [JohnnyCurran](https://elixirforum.com/u/JohnnyCurran) on Elixir Forums.  Maybe I'm way off base here but thought I'd open the PR.

Thanks!

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -276,12 +276,13 @@ defmodule Phoenix.HTML.Form do
 
   @doc false
   def fetch(%Form{errors: errors} = form, field) when is_atom(field) do
-    field_as_string = Atom.to_string(field)
+    field_as_string = field_to_string(field)
+    field_as_atom = field_to_atom(field)
 
     {:ok,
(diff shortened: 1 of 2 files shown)


josevalim:
This is trickier, as it means passing an invalid parameter dynamically will cause crashes when rendering the page due to the `to_existing_atom`. Historically speaking we do not convert strings to atoms in this project, which means that passing string keys means you can't access form data with atom values. So before merging this, I would like to hear the use cases for this feature, because it may not work as people expect.

josevalim:
Hi @sodapopcan! I went ahead and pushed a fix because this change required further consideration. Sorry for not being able to accept your PR at this time. :heart:

sodapopcan:
No problem at all!  I was just seeing if I could fix it and hope I didn't cause any PR-rejection stress 💚

PR #415, opened 2023-03-02
Fix links in docs for `Phoenix.HTML.FormData.to_form/2,4`

smaximov:


josevalim:
Thank you for finding it. I pushed a commit that explicitly documents the options and decouiples the two.

PR #422, opened 2023-05-05
Add support for lists of atoms

loganmac:
Original discussion: https://elixirforum.com/t/trouble-rendering-ecto-enum-array-in-liveview/55701

This PR should add support for rendering lists of atoms (i.e. if one of your schema's fields is an array of `Ecto.Enum`)

Added a test, tested in my own phoenix project and this fixes the error. 😄

### DIFF
--- a/lib/phoenix_html/safe.ex
+++ b/lib/phoenix_html/safe.ex
@@ -73,13 +73,17 @@ defimpl Phoenix.HTML.Safe, for: List do
     Phoenix.HTML.Engine.html_escape(h)
   end
 
+  def to_iodata(h) when is_atom(h) do
+    Phoenix.HTML.Engine.html_escape(Atom.to_string(h))
+  end
+
   def to_iodata({:safe, data}) do
     data
   end
 
   def to_iodata(other) do
     raise ArgumentError,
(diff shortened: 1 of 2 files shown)


josevalim:
I don't think this is useful because this would print your list of enums as: "danceswimyoga" which is hardly what you want in practice. You need to traverse the list yourself and convert it to a `ui+li` or similar in the HTML. :) Thank you!

loganmac:
Thanks @josevalim! Maybe this isn't the correct place to discuss this, but that's exactly what the generated code from `mix phx.gen.live MyContext MyModel mymodels services:array:string` does. I realize this isn't very useful, but it at least doesn't generate code with a broken test. I guess maybe a better contribution might be to fix the generators so they generate sensible code for those use-cases, what do you think? The same thing happens for maps too.

josevalim:
> I guess maybe a better contribution might be to fix the generators so they generate sensible code for those use-cases, what do you think? The same thing happens for maps too.

Yeah, that would be great. In those cases, a simpler solution may be to just wrap it in inspect: `<%= inspect field.value %>`. :)

PR #433, opened 2023-09-01
Update docs to note that form/1 is preferred over form_for/4

TheArrowsmith:
This was discussed in the Elixir Slack recently and I think it's worth making the docs clearer because it was a big point of confusion for me:

form_for_slack_conversation

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -402,7 +402,7 @@ defmodule Phoenix.HTML.Form do
 
   Note that if you are using Phoenix LiveView, `form_for/4` is no longer the
   preferred way to generate a form tag, and you should use
-  [Phoenix.Component.form/1](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#form/1)
+  [`Phoenix.Component.form/1`](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#form/1)
   instead.
 
   ## With changeset data
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -400,6 +400,11 @@ defmodule Phoenix.HTML.Form do
 
   We will explore all them below.
 
+  Note that if you are using Phoenix LiveView, `form_for/4` is no longer the
+  preferred way to generate a form tag, and you should use
+  [Phoenix.Component.form/1](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#form/1)
+  instead.
+
   ## With changeset data
 
   The entry point for defining forms in Phoenix is with


lib/phoenix_html/form.ex
```diff
@@ -400,6 +400,11 @@ defmodule Phoenix.HTML.Form do
 
   We will explore all them below.
 
+  Note that if you are using Phoenix LiveView, `form_for/4` is no longer the
+  preferred way to generate a form tag, and you should use
+  [Phoenix.Component.form/1](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#form/1)
```
josevalim, line 405:
```suggestion
  [`Phoenix.Component.form/1`](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#form/1)
```

josevalim:

lib/phoenix_html/form.ex
```diff
@@ -400,6 +400,11 @@ defmodule Phoenix.HTML.Form do
 
   We will explore all them below.
 
+  Note that if you are using Phoenix LiveView, `form_for/4` is no longer the
```
LostKobrakai, line 403: This could be confusing. One could be using phoenix 1.7 (which depends on LiveView for html), but not run any "LiveViews". They still want to use `<.form>`. 

Maybe `if you are using HEEx/function components …` would be closer? Both heex and function component mean the `phoenix_live_view` package is available, without the need of using "LiveViews".

TheArrowsmith: I agree that it's confusing, but it's worth noting that the language "if you are using Phoenix LiveView" already appears in this file on L348:

```elixir
  def form_for(form_data, action, options) when is_list(options) do
    IO.warn(
      "form_for/3 without an anonymous function is deprecated. " <>
        "If you are using Phoenix.LiveView, use the new Phoenix.Component.form/1 component"
    )

    %{Phoenix.HTML.FormData.to_form(form_data, options) | action: action}
  end
```

So if it's going to be changed, it should be changed in both places.

i-n-g-m-a-r:
Hi @TheArrowsmith 

According to your post here:
https://arrowsmithlabs.com/blog/you-can-stop-using-form-for-in-phoenix

"However, [since Phoenix LiveView v0.16.0](https://github.com/phoenixframework/phoenix_live_view/blob/v0.19.5/CHANGELOG.md#enhancements-27), there’s also the form/1 function component, which does essentially the same thing as form_for/4"

Following that link: "[Helpers] Introduce form function component which wraps Phoenix.HTML.form_for".

Then you state: "So that settles it: form_for/4 is deprecated and should no longer be used".

Now I'm confused, especially since `<.form :let={f} for={@changeset} ...` stopped working recently.

Fixing this using `FormData.to_form changeset, [ ... ]` works fine so I'm not sure if "deprecated" still applies.

SteffenDE:
> Now I'm confused, especially since <.form :let={f} for={@changeset} ... stopped working recently.

can you provide details? That should still work, although it’s not the recommended approach.

i-n-g-m-a-r:
Hi @SteffenDE right now I can't reproduce any problems with `for={@changeset}`. Not sure what happened. I've updated my dependencies to the latest versions a few times lately. It's hard to keep up.

Probably I messed up while refactoring, given that "should still work" suggests that particular functionality (accepting an Ecto.Changeset struct) was never changed.

Sorry about the confusion / false alarm.

However, I'm pretty happy actually using `Phoenix.HTML.FormData.to_form/2` because it gives me access to `:options` in `Phoenix.HTML.Form`. Where should I read about the (current) recommended approach?

SteffenDE:
The recommended approach is to use the `to_form` helper from `Phoenix.Component`: https://phoenix-live-view.hexdocs.pm/Phoenix.Component.html#to_form/2

It forwards the options to `Phoenix.HTML.FormData.to_form/2`.

PR #436, opened 2023-12-21
Require a data-to with the data-method before the click handler triggers a HTTP request 

tarzan:
We recently ran into a conflict with a third party js library that uses the `data-method` attribute on elements to signal which function should handle its events. The javascript bundled with `phoenix_html` also does a HTTP POST together with this non-sensical method attribute as a side effect.

The docs here might be misleading:
```
Support data-method="patch|post|put|delete" attributes, which sends the current click as a PATCH/POST/PUT/DELETE HTTP request. You will need to add data-to with the URL and data-csrf with the CSRF token value
```
as that would suggest that only valid HTTP request types are being handled, while instead every custom method results in a POST as well.

So I decided to restrict this to only trigger click handling when the `data-method` is indeed one of the aforementioned types.

### DIFF
--- a/priv/static/phoenix_html.js
+++ b/priv/static/phoenix_html.js
@@ -65,7 +65,7 @@
         return false;
       }
 
-      if (element.getAttribute("data-method")) {
+      if (element.getAttribute("data-method") && element.getAttribute("data-to")) {
         handleClick(element, e.metaKey || e.shiftKey);
         e.preventDefault();
         return false;


josevalim:
I think strictly speaking we allow any verb in "match" in Phoenix, not only these pre-defined ones. So we would need another way to address those conflicts. :(

tarzan:
> I think strictly speaking we allow any verb in "match" in Phoenix, not only these pre-defined ones. So we would need another way to address those conflicts. :(

Another option would be to require a `data-to` with the `data-method` as those two look like the bare minimum for constructing a HTTP request? 
Without the `data-to` the form is currently posting to `/current_path/NULL` which isn't ever wanted behaviour, or is it?

josevalim:
Yup, we need a data-to, so that would be a good fix.

tarzan:
> Yup, we need a data-to, so that would be a good fix.

Thanks! I redid the conditional to match on there being **both** a `data-method` and a `data-to` attribute before a HTTP request is being constructed.

josevalim:

PR #439, opened 2024-02-05
Add form action and consider input changed if action changes

chrismccord:
ref: todo

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -65,7 +65,6 @@ defmodule Phoenix.HTML.Form do
             name: nil,
             data: nil,
             action: nil,
-            method: nil,
             hidden: [],
             params: %{},
             errors: [],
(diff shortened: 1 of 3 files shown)


lib/phoenix_html/form.ex
```diff
@@ -202,7 +204,7 @@ defmodule Phoenix.HTML.Form do
         field
       )
       when is_atom(field) or is_binary(field) do
-    impl1 != impl2 or id1 != id2 or name1 != name2 or
+    impl1 != impl2 or id1 != id2 or name1 != name2 or form1.action != form2.action or
```
josevalim, line 207: Please pattern match on those as well for consistency?

lib/phoenix_html/form_data.ex
```diff
@@ -56,6 +56,8 @@ defprotocol Phoenix.HTML.FormData do
       applies if the field value is a list and no parameters were
       sent through the form.
 
+    * `:action` - The form action, such as the HTML `action` attribute
+      or LiveView action.
```
josevalim, line 60: Why the double meaning? Before it was always an atom.

lib/phoenix_html/form.ex
```diff
@@ -74,6 +75,7 @@ defmodule Phoenix.HTML.Form do
           source: Phoenix.HTML.FormData.t(),
           name: String.t(),
           data: %{field => term},
+          action: nil | atom() | String.t(),
```
josevalim, line 78:
```suggestion
          action: atom(),
```

PR #451, opened 2024-10-18
Add <hr/> support to options_for_select

gilbertbw:
This allows you to add separators to option lists, as described [here](https://developer.chrome.com/blog/hr-in-select).

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -297,6 +297,21 @@ defmodule Phoenix.HTML.Form do
       #=>   <option>France</option>
       #=> </optgroup>
 
+  Horizontal seperators can be added:
+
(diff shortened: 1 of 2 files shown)


josevalim (approved):

lib/phoenix_html/form.ex
```diff
@@ -297,6 +297,21 @@ defmodule Phoenix.HTML.Form do
       #=>   <option>France</option>
       #=> </optgroup>
 
+  Horizontal seperators can be added:
```
SteffenDE, line 300:
```suggestion
  Horizontal separators can be added:
```

SteffenDE (approved):

SteffenDE:
Thank you! 🙌

PR #452, opened 2024-10-21
add Phoenix.HTML.css_escape/1

SteffenDE:
See https://github.com/phoenixframework/phoenix_live_view/issues/3465 and https://github.com/philss/floki/pull/594.


lib/phoenix_html.ex
```diff
@@ -133,6 +133,21 @@ defmodule Phoenix.HTML do
...
+
+      iex> css_escape("hello world")
+      "hello\\\\ world"
+
+      iex> css_escape("-123")
+      "-\\\\31 23"
+
+  """
+  @spec css_escape(String.t()) :: String.t()
+  def css_escape(value), do: Phoenix.HTML.CSS.escape(value)
```
josevalim, line 149: We can just inline it here IMO, no need of a separate module.

josevalim (approved):

PR #456, opened 2025-01-22
Allow keyword list options to use nil as key and/or value 

### LINKED ISSUE #455: `nil` handled differently for select options depending on "format"
```
17:50:51.170244 <0.728.0> Phoenix.HTML.Form.options_for_select([nil], nil)
{:safe,
 [[], 60, "option", [32, "value", 61, 34, "", 34], 62, "", 60, 47, "option", 62]}
iex(2)> Phoenix.HTML.Form.options_for_select([[key: nil, value: nil]], nil)
** (ArgumentError) expected :key key when building <option> from keyword list: [value: nil]
    (phoenix_html 4.1.1) lib/phoenix_html/form.ex:320: anonymous fn/3 in Phoenix.HTML.Form.escaped_options_for_select/2
    (elixir 1.18.1) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
    (phoenix_html 4.1.1) lib/phoenix_html/form.ex:305: Phoenix.HTML.Form.options_for_select/2
    iex:2: (file)
```

I just changed the former to the latter (to add a disabled key) and this broke my select option building. I had to use `[key: "#{val}", value: "#{val}"]` as a workaround

LostKobrakai:
Fixes #455

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -334,18 +334,25 @@ defmodule Phoenix.HTML.Form do
         [acc | option(option_key, option_value, [], selected_values)]
 
       options, acc when is_list(options) ->
-        if not Keyword.has_key?(options, :key) do
-          raise ArgumentError,
-                "expected :key key when building <option> from keyword list: #{inspect(options)}"
-        end
+        {option_key, options} =
+          case List.keytake(options, :key, 0) do
+            nil ->
+              raise ArgumentError,
+                    "expected :key key when building <option> from keyword list: #{inspect(options)}"
 
-        if not Keyword.has_key?(options, :value) do
-          raise ArgumentError,
-                "expected :value key when building <option> from keyword list: #{inspect(options)}"
-        end
+            {{:key, key}, options} ->
+              {key, options}
+          end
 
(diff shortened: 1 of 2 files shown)


lib/phoenix_html/form.ex
```diff
@@ -334,17 +334,18 @@ defmodule Phoenix.HTML.Form do
...
-        option_key ||
+        if not Keyword.has_key?(options, :key) do
           raise ArgumentError,
                 "expected :key key when building <option> from keyword list: #{inspect(options)}"
+        end
 
-        {option_value, options} = Keyword.pop(options, :value)
-
-        option_value ||
+        if not Keyword.has_key?(options, :value) do
```
josevalim, line 342: If you want to do a single pass, you can use `List.keytake`. Thoughts?

LostKobrakai: Ah nice. I only looked at Keyword apis.

josevalim: Ok, please use keytake and I will merge it once ready! Thanks!

josevalim:

PR #460, opened 2025-02-21
Expand documentation of options_for_select

ponychicken:
See https://github.com/phoenixframework/phoenix/pull/5764

### DIFF
--- a/lib/phoenix_html/form.ex
+++ b/lib/phoenix_html/form.ex
@@ -273,10 +273,25 @@ defmodule Phoenix.HTML.Form do
   end
 
   @doc """
-  Returns options to be used inside a select.
+  Returns options to be used inside a select element.
 
-  This is useful when building the select by hand.
-  It expects all options and one or more select values.
+  `options` is expected to be an enumerable which will be used to
+  generate each `option` element. The function supports different data
+  for the individual elements:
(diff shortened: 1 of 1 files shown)


lib/phoenix_html/form.ex
```diff
@@ -273,10 +273,25 @@ defmodule Phoenix.HTML.Form do
...
+    * keyword lists - each keyword list is expected to have the keys
+      `:key` and `:value`. Additional keys such as `:disabled` may
+      be given to customize the option.
+    * two-item tuples - where the first element is an atom, string or
+      integer to be used as the option label and the second element is
+      an atom, string or integer to be used as the option value
+    * simple atom, string or integer - which will be used as both label and value
+      for the generated select
+    
+  ## Option groups
```
josevalim, line 290:
```suggestion
  for the individual elements:

    * keyword lists - each keyword list is expected to have the keys
      `:key` and `:value`. Additional keys such as `:disabled` may
      be given to customize the option.
    * two-item tuples - where the first element is an atom, string or
      integer to be used as the option label and the second element is
      an atom, string or integer to be used as the option value
    * simple atom, string or integer - which will be used as both label and value
      for the generated select
    
  ## Option groups

```

josevalim:

PR #463, opened 2025-04-30
Add `Phoenix.HTML.Safe` to Duration

joshua-bouv:
I was creating a form which had a set of options, the values being durations

```elixir
<.input
  label="Duration"
  field={@form[:duration]}
  type="select"
  options={[
    {"Daily", Duration.new!(day: 1)},
    {"Weekly", Duration.new!(week: 1)},
    {"Monthly", Duration.new!(month: 1)},
    {"Yearly", Duration.new!(year: 1)}
  ]}
/>
```

`Duration` has `to_iso8601` and `from_iso8601` functions so I assumed this would work like `Date` and `Time` does (auto converts to ISO), but `Phoenix.HTML.Safe` hadn't been implemented for Duration causing this error: `protocol Phoenix.HTML.Safe not implemented for type Duration (a struct).`

This PR adds the `Phoenix.HTML.Safe` implentation. Their might of been a reason why this wasn't added originally though, but im not too sure why that could be?

### DIFF
--- a/lib/phoenix_html/safe.ex
+++ b/lib/phoenix_html/safe.ex
@@ -44,8 +44,10 @@ defimpl Phoenix.HTML.Safe, for: DateTime do
   end
 end
 
-defimpl Phoenix.HTML.Safe, for: Duration do
-  defdelegate to_iodata(data), to: Duration, as: :to_iso8601
+if Code.ensure_loaded?(Duration) do
+  defimpl Phoenix.HTML.Safe, for: Duration do
+    defdelegate to_iodata(data), to: Duration, as: :to_iso8601
+  end
 end
 
 defimpl Phoenix.HTML.Safe, for: List do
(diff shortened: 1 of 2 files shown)


SteffenDE:
The minimum supported Elixir version is 1.7 at the moment https://github.com/phoenixframework/phoenix_html/blob/c11a9e3b87fa247b31a8ccf1265d6588e1ddaa04/mix.exs#L12
which does not know anything about Duration, so we might need to conditionally define this? @josevalim

joshua-bouv:
> The minimum supported Elixir version is 1.7 at the moment
> 
> https://github.com/phoenixframework/phoenix_html/blob/c11a9e3b87fa247b31a8ccf1265d6588e1ddaa04/mix.exs#L12
> 
> which does not know anything about Duration, so we might need to conditionally define this? @josevalim

Good spot!

If the decision is to add`if Version.match?(System.version(), ">= 1.17.0") do` around the function (and test) I have the commit on stand-by 👍

josevalim:
@joshua-bouv yeah, you will need to wrap both the definition and the tests in `if Code.ensure_loaded?(Duration) do` blocks. :)

joshua-bouv:
Done 👍

josevalim:

